PR #15, opened 2019-01-09
Change companero to capitan; use package_release: stable/

brandones:


install.sh
```diff
@@ -37,7 +37,7 @@ else
 fi 
 
 # For Haiti servers, don't update the tzdata package
-if [$HOSTNAME == ""]
+if [ "$HOSTNAME" = "" ]
```
brandones: Fixed this bash syntax error.

PR #18, opened 2019-03-19
MEX-204 Mexico servers should install with the correct tomcat7 settings and apache disabled

brandones:
https://tickets.pih-emr.org/browse/MEX-204

### DIFF
--- a/hieradata/ces-capitan.yaml
+++ b/hieradata/ces-capitan.yaml
@@ -6,4 +6,5 @@ mysql_root_password: ENC[oWqIfmn/ER4CYIvjJSpIzQ==]
 mysql_bind_address: 0.0.0.0
 openmrs_db_password: ENC[3RqMvut63WqWc8IDs/GOnQ==]
 mirth_password: ENC[rtk/p1SzQjT3zkMJx7DJ0g==]
+tomcat_enable_http_8080: true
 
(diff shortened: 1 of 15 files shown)


brandones:
Oh cool, I didn't know about template files. Preference between ERB and EPP?

I had removed apache2 from `site.pp`, I guess it got lost on the way into git. Updated.

brandones:
Never mind; seems clear from the docs that EPP is what we want (their ERB templater doesn't support parameterization).

brandones:
Oh wow, don't merge this, it seems quite broken

brandones:
Ah never mind, I think it's fine actually. My git situation is a little messy right now.

mogoodrich:
So.... long answer... I hadn't heard of epp before and it looks like that is because when we originally installed puppet it was at 3.1.0 and epp didn't exist... we want to update puppet to 4.x (and could likely do that at part of our update to Ubuntu 16.04) but right now most of servers are running 3.1.0 and it looks like epp will break there:

https://puppet.com/docs/puppet/5.3/lang_template_epp.html

Any chance we could do what we need to do with erb for now? :)

brandones:
ERB doesn't accept template parameters, so I'm not sure what the point would be. NB, my server has been running Puppet 4.10 and everything has worked just fine.

mogoodrich:
yeah, I don't have a problem with upgrading to 4.1.0, but if I merge this in, assumedly every server running Puppet 3.10 will crap out until we get the update. I could be missing something but  the $enable_http_8080 should be available in the erb... take a look at feature_toggles.erb or any of the other erb files in the mirebalais-modules folder... I don't see an example of an if/then, only a direct replacement, but I think we did it somewhere, I just couldn't find it.

brandones:
Oh cool, looks like I had misread the Puppet docs. Yeah I'll go ahead and change that over to ERB.

mogoodrich:
Cool, this looks good now, thanks... are we ready to merge on your end?

PR #29, opened 2019-04-30
LIB: Added JJ Dossen as site/facility supported, added config files

alaboso:
- Added JJ Dossen as site/facility supported
- Added config files


alaboso:
I have changed so that we have __liberia-harper__  then;

- __liberia-harper-kouka__
- __liberia-harper-jjdossen__

They use same location, same site but different set of configs.

mogoodrich:
Looks good, merging!

PR #36, opened 2019-06-18
MEX-362 Height/weight graph on main patient dashboard

brandones:
https://pihemr.atlassian.net/browse/MEX-362

### DIFF
--- a/mirebalais-modules/openmrs/files/config/pih-config-mexico.json
+++ b/mirebalais-modules/openmrs/files/config/pih-config-mexico.json
@@ -8,6 +8,7 @@
     "ancProgram",
     "asthmaProgram",
     "chwApp",
+    "bmiOnClinicianDashboard",
     "clinicianDashboard",
     "dataExports",
     "diabetesProgram",


mirebalais-modules/openmrs/files/config/pih-config-mexico.json
```diff
@@ -8,6 +8,7 @@
     "ancProgram",
     "asthmaProgram",
     "chwApp",
+    "bmiOnClinicianDashboard",
```
mogoodrich: I'm wondering if we want to make this a specific boolean config parameter elsewhere in the pih-config rather than a component, since it's such a small, targeted feature, not really a component.  Thoughts?

brandones: Sure, though not sure where it'd go in the pih-config. Thoughts?

mogoodrich: Maybe a new "clinicianDashboard" sectionwith a parameter "showBMI"?

```
clinicianDashboardConfig: {
   showBMI: true
}
```

Then we should potentially move other configurations for the clinician dashboard here as well... 

Thought that's just a off-the-cuff answer, I'd also be up for defining an overall strategy of how we want to organize things in our config, as we've been doing the design rather ad hoc so far...

brandones: This would be a new paradigm for us, whereas everything else in CALF is based on components. I'd be into a new paradigm, but yeah, I like the idea of first developing the vision / overall strategy that it would be aiming towards.

One possible dream might be configs like

```
clinicainDashboardConfig: {
  firstColumn: [
    'coreapps.clinicianfacing.visits',
    'pih.app.bmiGraph'
  ],
  secondColumn: [
    'pih.app.lastVitals'
  ]
},
visitActions: [
  'vitals',
  'primaryCare'
]
```

The job of CALF then becomes to initialize all our custom apps and actions and add them to extensions in the order specified by the config. Components can mostly be done away with. As can all the site and country branching in CALF. All that brittle ordering stuff in CALC as well.

An elaboration/improvement might be

```
clinicainDashboardConfig: {
  firstColumn: [
    { id: 'pih.app.bmiGraph'
      maxResults: 12
    },
    ...
```

brandones: Wait, I actually really like this proposal. This would give us type-checked UUIDs, easy reusability of components, and a high level of configuration without messing with Java. Also, while non-trivial, I don't think it would be crazy difficult. And could be broken up into separate chunks of work -- first the dashboards, then the visit actions. If it doesn't seem unreasonable to you, can we talk about it more sometime soon?

mogoodrich: This absolutely seems reasonable to me, and I don't think we really have to do all at once....  sure, we can talk more...  (thoiugh I think you're out tomorrow, and then I'm out until July 8th...)

brandones:
Per comment on https://github.com/PIH/openmrs-module-mirebalais/pull/54#pullrequestreview-255311104, I'll merge and then try to make sure we fix this sooner rather than later.

PR #42, opened 2019-07-25
Mex 399 ces node

jmbabazi:
@mogoodrich and @brandones , when you get a chance. Would you please review this? In addition to the mexico node, I added the webapp name so that apache can pick it up.

### DIFF
--- a/hieradata/ces-honduras.yaml
+++ b/hieradata/ces-honduras.yaml
@@ -1,4 +1,5 @@
 ---
+webapp_name: openmrs
 package_release: 'stable/'
 pih_config: mexico,mexico-honduras
 config_dir: mexico
--- a/hieradata/ces.pih-emr.org.yaml
+++ b/hieradata/ces.pih-emr.org.yaml
@@ -1,4 +1,5 @@
 ---
+webapp_name: openmrs
 package_release: 'stable/'
 site_domain: ces.pih-emr.org
 pih_config: mexico,mexico-production
--- a/hieradata/ces.pih-emr.org.yaml
+++ b/hieradata/ces.pih-emr.org.yaml
@@ -4,8 +4,9 @@ site_domain: ces.pih-emr.org
 pih_config: mexico,mexico-production
 config_dir: mexico
 
+server_timezone: America/Mexico_City
 mysql_root_password: ENC[oWqIfmn/ER4CYIvjJSpIzQ==]
-mysql_bind_address: 0.0.0.0
+mysql_bind_address: 127.0.0.1
 openmrs_db_password: ENC[3RqMvut63WqWc8IDs/GOnQ==]
 mirth_password: ENC[rtk/p1SzQjT3zkMJx7DJ0g==]
 
--- a/manifests/site.pp
+++ b/manifests/site.pp
@@ -240,6 +240,33 @@ node 'wellbody.pih-emr.org' {
   include crashplan
 }
 
+node 'ces.pih-emr.org' {
+
+  class { 'apt':
+    always_apt_update => true,
+  }
+
+  include security
+  include mailx
+  include ntpdate
+  include apt_upgrades
+  include wget
+
+  include java
+  include mysql_setup
+  include apache2
+  include tomcat
+
+  include openmrs
+  include openmrs::initial_setup
+
+  #include monitoring
+  include logging
+
+  include openmrs::backup
+  include crashplan
+}
+
 node 'kouka.pih-emr.org', 'padi.pih-emr.org', 'ci.pih-emr.org', 'ami.pih-emr.org', 'lespwa.pih-emr.org' {
 
   class { 'apt':
--- a/manifests/site.pp
+++ b/manifests/site.pp
@@ -240,33 +240,6 @@ node 'wellbody.pih-emr.org' {
   include crashplan
 }
 
-node 'ces.pih-emr.org' {
-
-  class { 'apt':
-    always_apt_update => true,
-  }
-
-  include security
-  include mailx
-  include ntpdate
-  include apt_upgrades
-  include wget
-
-  include java
-  include mysql_setup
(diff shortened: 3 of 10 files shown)


hieradata/ces-honduras.yaml
```diff
@@ -1,4 +1,5 @@
 ---
+webapp_name: openmrs
```
mogoodrich: the common.yaml file sets the webapp_name to "openmrs" and if a variable is not found in site yaml file it should default to the one in the common.yaml, so I don't think all these are needed... were you noticing a problem with this @jmbabazi ?

manifests/site.pp
```diff
@@ -260,10 +260,12 @@
   include openmrs
   include openmrs::initial_setup
 
+  include openmrs::backup
```
mogoodrich: I think this is the node for all the staging servers, so we weren't worrying about backing those us?

hieradata/ces.pih-emr.org.yaml
```diff
@@ -1,11 +1,13 @@
...
+webapp_name: openmrs
 package_release: 'stable/'
 site_domain: ces.pih-emr.org
 pih_config: mexico,mexico-production
 config_dir: mexico
 
+server_timezone: America/Mexico_City
 mysql_root_password: ENC[oWqIfmn/ER4CYIvjJSpIzQ==]
-mysql_bind_address: 0.0.0.0
+mysql_bind_address: 127.0.0.1
```
mogoodrich: Like webapp_name we should be able to just remove the mysql_bind_address here since the common.yaml sets it to "127.0.0.1".

It does make sense to change this to 127.0.0.1 to lock it down.  (We may need to open it up at some point when we get smoke tests running against ces.pih-emr.org, but we can cross that bridge when we come to it)

We also should remove the mysql_bind_address: 0.0.0.0 for all the other CES servers, unless there's a reason we are doing this I don't know about?

See:
https://pihemr.atlassian.net/browse/UHM-3075

mirebalais-modules/tomcat/templates/server.xml.erb
```diff
@@ -93,7 +93,7 @@
     -->
 
     <!-- Define an AJP 1.3 Connector on port 8009 -->
-    <Connector port="8009" address="127.0.0.1" protocol="AJP/1.3" redirectPort="8443"
+    <Connector port="8009" address="localhost" protocol="AJP/1.3" redirectPort="8443"
```
mogoodrich: Why is this changed?  Is this to fix that problem we had when one of the servers was recognizing 127.0.0.1 as the localhost ip? (I tried to find the ticket on this to remind myself, but couldn't...)

hieradata/ces.pih-emr.org.yaml
brandones: Fine with me!

manifests/site.pp
```diff
@@ -260,10 +260,12 @@
...
+
   #include monitoring
 }
 
-node 'ces.pih-emr.org' {
+node 'ces.pih-emr.org', 'ces-capitan.pih-emr.org', 'ces-laguna', 'ces-honduras', 'ces-letrero', 'ces-matazano', 'ces-monterrey', 'ces-plan', 'ces-reforma', 'ces-salvador', 'ces-soledad'    {
```
brandones: CES needs three different kinds of nodes:

`ces.pih-emr.org` should be the production node. It should be configured like all our other production instances.

A node with a name like `ces-ci.pih-emr.org` or `ces-demo.pih-emr.org` should be the CI node. I added a YAML file `hieradata/ces-demo.pih-emr.org.yaml`. You can change it to `ces-ci` if you prefer. This node should probably be added to line 243 of this file, up with `kouka.pih-emr.org', 'padi.pih-emr.org', 'ci.pih-emr.org'` etc.

Nodes with names like `ces-capitan`, `ces-honduras`, etc. These are defined below on line 289/291. You shouldn't need to mess with them right now.

So, regarding this line:
- `ces.pih-emr.org` should probably just be added to whatever production node configuration
- `ces-capitan.pih-emr.org` should not exist
- `ces-laguna` etc should not be added here -- they should be left as they are on line 289/291.

hieradata/ces-honduras.yaml
jmbabazi: Yes, I was noticing a problem. common.yaml does not replace the webapp_name in https://github.com/PIH/mirebalais-puppet/blob/90c82c44d2ea34efb0a5c96dc6b11dd8ef015ac9/mirebalais-modules/apache2/templates/default-ssl.conf.erb.

mirebalais-modules/tomcat/templates/server.xml.erb
jmbabazi: It really doesn't matter whichever way you have it as long as everything is mapped correctly in /etc/hosts. My thinking was that we need to standardize as in https://github.com/PIH/mirebalais-puppet/blob/master/mirebalais-modules/apache2/templates/workers.properties.erb

hieradata/ces.pih-emr.org.yaml
jmbabazi: If you remove webapp_name, you will have to add the JKMount manually in default-ssl. as for the bind address I defer to brandon.

hieradata/ces-honduras.yaml
mogoodrich: @jmbabazi this is very strange, can you probe another detail?

mirebalais-modules/tomcat/templates/server.xml.erb
mogoodrich: Cool, makes sense!

jmbabazi:
This won't be merged in. I'll create another ticket for solving the defaults-ssl issues.
The issues are on JkMount and ssl files.

PR #46, opened 2019-08-06
MEX-425 Strange address hierarchy behavior when not all levels present

brandones:
https://pihemr.atlassian.net/browse/MEX-425


mogoodrich:
@brandones what's the change here?  Adding in the "---" instead of empty rows?

brandones:
Yep, that's it.

mogoodrich:
LGTM

PR #50, opened 2019-09-25
MEX-429 Add documentation to puppet install

brandones:
https://pihemr.atlassian.net/browse/MEX-429


manifests/site.pp
```diff
@@ -306,6 +306,8 @@
   include openmrs
   include openmrs::initial_setup
 
+  include emr_user_resources
```
mogoodrich: Since this is very targeted to CES laptops, probably better to use a less-generic name?

brandones: Sure, fixed

PR #56, opened 2019-10-16
UHM-4112 Mexico and Liberia production servers should automatically run liquibase updates

brandones:
Just sets openmrs_auto_update_database to true in the hieradata files.

https://pihemr.atlassian.net/browse/UHM-4112


brandones:
Updated to remove the property from ces-ci, on Mark's advice that liquibase runs be left to run manually on CI servers.

alaboso:
@brandones Kindly, add the same to jjdossen.pih-emr.org.yaml. It's for JJD.
Otherwise, all looks good.

Thanks,

brandones:
Thanks Mark and Amos!

PR #57, opened 2019-10-21
MEX-129 Make session timeout longer

brandones:
https://pihemr.atlassian.net/browse/MEX-129

### DIFF
--- a/mirebalais-modules/openmrs/manifests/init.pp
+++ b/mirebalais-modules/openmrs/manifests/init.pp
@@ -7,12 +7,14 @@ class openmrs (
     $package_version = hiera('package_version'),
     $webapp_name = hiera('webapp_name'),
     $tomcat = hiera('tomcat'),
+    $tomcat_webapp_dir = hiera('tomcat_webapp_dir'),
     $junit_username = hiera('junit_username'),
     $junit_password = decrypt(hiera('junit_password')),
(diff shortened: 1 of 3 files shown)


mirebalais-modules/openmrs/manifests/init.pp
```diff
@@ -101,6 +103,29 @@
     require => File["/home/${tomcat}/.OpenMRS"]
   }
 
+  file_line { "web_xml_session_config":
+    ensure => "present",
+    path => "${tomcat_webapp_dir}/openmrs/WEB-INF/web.xml",
```
mogoodrich: Annoyingly the Mirebalais app has a different webapp name (in hindsight we should have just used openmrs!!) so instead of hardcoding "openmrs" you'll need to use the webapp_name yml variable here (and in the ones below)

mirebalais-modules/openmrs/manifests/init.pp
```diff
@@ -101,6 +103,29 @@
...
+    line => "        <session-config>",
+    match => "\s*<session-config>"
```
mogoodrich: all three of these will need a require => "Package['pihemr']" to ensure that they run *after* the PIH EMR package is installed.  

(Will be bit hard to test this... because just because the package has been installed, not sure if that ensures that the web.xml file will be installed)

PR #60, opened 2019-10-30
Components for graphs

jmbabazi:


mogoodrich:
@jmbabazi I don't think we actually need to add these as components here?  I think all that is needed is to change the enableMentalHealth method in the CALF to use the graphfactory, which just provides some reusable shortcuts for creating graphs.

brandones:
What's up with this PR? Can we close this, or does it address some yet-unmet goals?

mogoodrich:
Thanks @brandones ... I think we ended up handling this via another method though I don't recall 100% @jmbabazi ?

I will close this out, but feel free to reopen if I've missed anything.

PR #63, opened 2019-11-07
Upgrading puppet stdlib using puppet module

jmbabazi:


gem-update.sh
```diff
@@ -31,5 +31,7 @@ librarian-puppet install
 
 fi
 
+puppet module upgrade puppetlabs-stdlib --version 6.1.0 --modulepath=/etc/puppet/modules
```
brandones: Based on [the compatibility sheet](https://forge.puppet.com/puppetlabs/stdlib/compatibility), this version of stdlib shouldn't be compatible with puppet below version  5.5.10. What's the version of puppet you checked it with? Are you able to try this on a computer with puppet 3.x applied?

jmbabazi: Thanks @brandones , this seems to work on puppet 4.10.10 which is what we use on Ubuntu 16. I think I am going to change to version 4.25.1(which still doesn't work on 14.04 but the overall idea is to completely move away from ubuntu 14 and puppet 3...)

brandones:
Merge whenever, and then please ping me so that I can get https://pihemr.atlassian.net/browse/MEX-129 back in.

Puppetfile.lock
```diff
@@ -2,7 +2,7 @@ FORGE
   remote: https://forgeapi.puppetlabs.com
   specs:
     maestrodev-wget (1.7.0)
-    puppetlabs-stdlib (4.12.0)
+    puppetlabs-stdlib (4.25.1)
     puppetlabs-vcsrepo (1.3.1)
```
mogoodrich: I think technically we are supposed to update the Puppetfile as well, but not sure.. but, yeah, let's just go ahead with this for now.

gem-update.sh
```diff
@@ -31,5 +31,8 @@ librarian-puppet install
 
 fi
 
+#this is include in Puppetfile.lock
+#puppet module upgrade puppetlabs-stdlib --version 6.1.0 --modulepath=/etc/puppet/modules
+
```
mogoodrich: let's just delete this, as we will forget what is does in the future!

PR #64, opened 2019-11-08
MEX-459 "Confirmed" checkbox in Diagnosis wiget should instead say "First time" ("1a vez")

brandones:
https://pihemr.atlassian.net/browse/MEX-459

Initializer pulls in the message property from the `configuration/messageproperties` directory. Anything defined there become the top-priority message. These configurations correspond to a single country. Thus this allows us to override a message property just for Mexico.


mogoodrich:
Do you want overwrite the English code as well?
Also, I think messages.properties files support comments, so I'd add a quick note as to why we are doing this for reference.

mogoodrich:
(But beyond that LGTM)

brandones:
Good ideas, thanks @mogoodrich !

PR #68, opened 2019-12-19
Changing percona password

jmbabazi:


brandones:
@jmbabazi It might be more sensible to have someone who understands this at all review this, but I'm interested in learning what's going on here, if you have the time to explain -- how do we use Percona, and what's going on in this change?

mirebalais-modules/mirebalais_reporting/templates/mirebalais-percona-restore.sh.erb
```diff
@@ -45,7 +45,7 @@ perconaRestore () {
                 # We should be stopping mysql first
                 service mysql stop
                 rm -rf ${dataDir}/*
-                innobackupex  --defaults-file="/etc/mysql/my.cnf"  --copy-back ${perconaRestoreDir} &>> ${perconaLogs}/`date +%y%m%d%H%M%S`_perconarestore.log
```
mogoodrich: What does this change do?

mirebalais-modules/mirebalais_reporting/templates/mirebalais-production-percona-backup.sh.erb
```diff
@@ -5,7 +5,7 @@ set -x
 backupDir="/home/percona/backups"
 openmrsDB="openmrs"
 perconaUser="percona"
-perconaPassword='<%= @mysql_root_password %>'
+perconaPassword='<%= @backup_password %>'
```
mogoodrich: Has the password for the Percona user been changed?  Is this a mysql account called "percona" that is used to dump the database on the production server?  

If this account is needed on the production server, I would recommend adding code to production_setup.pp to set that user up.  If this account is needed on the reporting server, I would recommend adding code to reporting_setup.pp to that user up there.  This way you can ensure that the password for that user is what you expect.

There are specific Puppet modules that can be used to get up MySQL users and grant them privileges on databases.  You can see an example of how we use it to create the "openmrs" user here and grant privileges to that user to the OpenMRS database.

https://github.com/PIH/mirebalais-puppet/blob/master/mirebalais-modules/openmrs/manifests/initial_setup.pp#L14

https://github.com/PIH/mirebalais-puppet/blob/master/mirebalais-modules/openmrs/manifests/initial_setup.pp#L20

You could do something similar in production_setup.pp and reporting_setup.pp as needed.

mirebalais-modules/mirebalais_reporting/templates/mirebalais-percona-restore.sh.erb
jmbabazi: As percona versions changes there are changes to the percona commands too.. "--defaults-file="  was an option used in the last version that we were using. More about the --default-file argument can be found at https://www.percona.com/doc/percona-xtrabackup/2.2/xtrabackup_bin/xbk_option_reference.html.

I'm not sure if you saw on telegram but my suggestion was to disable the re-installation of percona every time we do a release on mirebalais because the underlying cause of the reporting server failing is the version difference of percona on both the prod and report servers.

mirebalais-modules/mirebalais_reporting/templates/mirebalais-production-percona-backup.sh.erb
jmbabazi: This user is not need on the reporting server since he is copied over from the prod server. There was a password mismatch, This should now solve the issue.

> mirebalais-production-percona-backup
This can be removed. 

> * The "mirebalais-percona-backup" should likely live in the "mirebalais_reporting" directory, not "openmrs", since that is where it is referenced.

It should live in openmrs because this is the script that creates the backup in mirebalais_reporting there is a script called mirebalais-percona-restore.sh.erb which does the backup restoration.

mirebalais-modules/mirebalais_reporting/templates/mirebalais-percona-backup.sh.erb
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
```
mogoodrich: Did the version of this file in openmrs/templates/ get removed?  I can't quite tell from this PR?

mirebalais-modules/mirebalais_reporting/templates/mirebalais-production-percona-backup.sh.erb
mogoodrich: Should we be deleting mirebalais-production-percona-backup then?

mogoodrich:
@jmbabazi it looks like from the PR "mirebalais-production-percona-backup" has not been completed?

PR #69, opened 2020-01-08
MEX 65: Make registration lead to clinical dashboard

brandones:
https://pihemr.atlassian.net/browse/MEX-65

### DIFF
--- a/mirebalais-modules/openmrs/files/config/pih-config-mexico.json
+++ b/mirebalais-modules/openmrs/files/config/pih-config-mexico.json
@@ -41,7 +41,7 @@
   "registrationConfig": {
     "allowUnknownPatients": false,
     "allowManualEntryOfPrimaryIdentifier": false,
-    "afterCreatedUrl": "mirebalais/patientRegistration/afterRegistration.page?patientId={{patientId}}&encounterId={{encounterId}}",
+    "afterCreatedUrl": "/coreapps/clinicianfacing/patient.page?patientId={{patientId}}",
     "similarPatientsSearch": {
       "addressFields": {
         "countyDistrict": "1",


mirebalais-modules/openmrs/files/config/pih-config-mexico.json
```diff
@@ -31,8 +31,8 @@
   ],
   "scheduleBackupReports": "false",
   "dashboardUrl": "/coreapps/clinicianfacing/patient.page?patientId={{patientId}}&app=pih.app.clinicianDashboard",
-  "visitsPageUrl": "/pihcore/visit/visit.page?patient={{patient.uuid}}#/visitList",
-  "visitsPageWithSpecificUrl": "/pihcore/visit/visit.page?patient={{patient.uuid}}&visit={{visit.uuid}}#/overview",
+  "visitsPageUrl": "/coreapps/patientdashboard/patientDashboard.page?patientId={{patientId}}",
+  "visitsPageWithSpecificUrl": "/coreapps/patientdashboard/patientDashboard.page?patientId={{patientId}}&visitId={{visitId}}",
```
brandones: These didn't seem to change anything. Looking at CALF, it looks like they're not used?

mogoodrich: Ah, this is because Mexico uses the old visits page? Makes sense.

(I will follow up to see if I can remember where, if anywhere, this is used)

mogoodrich: Looks like they are used here to set a GP:

https://github.com/PIH/openmrs-module-mirebalais/blob/master/api/src/main/java/org/openmrs/module/mirebalais/MirebalaisHospitalActivator.java#L130

And then this global property is used in tne "Visit Sections" fragment defined by coreapps:

https://github.com/openmrs/openmrs-module-coreapps/blob/master/omod/src/main/java/org/openmrs/module/coreapps/fragment/controller/clinicianfacing/VisitsSectionFragmentController.java#L90

However, this would make me think that the links on the "Recent Visits" clinician dashboard widget would be going to the wrong location before this fix, and it looks to be working... I'm likely missing something.

brandones: Hmm, I'm going to undo this change for now.

PR #70, opened 2020-01-09
PER-334: Set up site for Peru

brandones:
https://pihemr.atlassian.net/browse/PER-334

Depends on https://github.com/PIH/openmrs-module-pihcore/pull/137

### DIFF
--- a/hieradata/peru-ci.pih-emr.org.yaml
+++ b/hieradata/peru-ci.pih-emr.org.yaml
@@ -1,0 +1,11 @@
+---
+package_release: 'unstable/'
+site_domain: peru-ci.pih-emr.org
+pih_config: peru
+config_dir: peru
+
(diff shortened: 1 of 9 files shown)


brandones:
Why? I've never edited the dispositions-mexico file. Someone copy-pasted it from somewhere else and now if we want to get rid of all that mostly-identical boilerplate we have to go through and manually verify that it's actually identical.

mogoodrich:
@brandones that's fair... I'd be fine with renaming dispositions-mexico to disposition-default and using it everywhere where it's the same... I just think it's risky to have a file "dispositions-mexico" used in another country because if we *did* decide we needed to change the dispositions for Mexico at some point someone could assume that changing "dispositions-mexico" would only affect Mexico.

brandones:
Good call, fixed!

PR #74, opened 2020-01-15
Add SPA to Puppet

brandones:


mirebalais-modules/openmrs/manifests/init.pp
```diff
@@ -35,6 +35,7 @@
 
   include openmrs::owa
   include openmrs::pwa
+  include openmrs::spa
```
mogoodrich: Maybe for starters we just include this in node.pp for the Mexico node to test?

mirebalais-modules/nodejs/manifests/init.pp
```diff
@@ -0,0 +1,18 @@
...
+class nodejs() {
+
+  package { 'curl':
+    ensure => installed,
+  }
+
+  exec { 'download node installer':
+    command => 'curl -sL https://deb.nodesource.com/setup_12.x | sudo -E bash -',
+    require => Package['curl']
+  }
```
mogoodrich: There's no apt-get install of node, or are you trying to enforce a specific version here?

mirebalais-modules/openmrs/manifests/spa.pp
```diff
@@ -0,0 +1,25 @@
+class openmrs::spa (
+  $tomcat = hiera('tomcat')
+) {
+
+  include nodejs
+
+  package { 'git':
+    ensure => installed,
+  }
```
mogoodrich: Annoyingly, this might cause a clash if there's anothter "package git" I believe?

mirebalais-modules/openmrs/manifests/spa.pp
```diff
@@ -0,0 +1,25 @@
...
+  vcsrepo { "/home/${tomcat}/.OpenMRS/frontend-build":
+    ensure   => latest,
+    provider => git,
+    source   => 'https://github.com/pih/spa-frontend.git',
+    revision => 'master',
+    require => [Service[$tomcat],Package['git']]
+  }
+
+  exec{'build_and_link_spa':
+    command => "npm install && npm run build && ln -sf $(realpath openmrs/frontend /home/${tomcat}/.OpenMRS/frontend)",
```
mogoodrich: It is generally best practice to install Node and build locally, or would it be better for us to install a published artifact?

mseaton: Hmm.  I'd tend to agree.  Shouldn't our spa be pre-packaged and deployed?  We don't pull modules down from maven in our puppet script - they are packaged on CI (and that bundle is deployed consistenly everywhere as a part of that package).

mirebalais-modules/nodejs/manifests/init.pp
brandones: Yeah, unfortunately the apt node is wayyyy out of date. This is the recommended way of installing.

mirebalais-modules/openmrs/manifests/spa.pp
brandones: Oh that's right, and there is another package git -- I had deleted it when I was testing this, because I thought it was superfluous, and then I added it back. Lemme fix this.

mirebalais-modules/openmrs/manifests/spa.pp
brandones: Oh yeah, that sounds like the right way to go, actually. I'm opening a discussion [on Slack](https://openmrs.slack.com/archives/CHP5QAE5R/p1579128063025000).

mogoodrich: Great, thanks!

mirebalais-modules/openmrs/manifests/init.pp
brandones: Where is the node.pp?

mogoodrich: https://github.com/PIH/mirebalais-puppet/blob/master/manifests/site.pp

mogoodrich: (Sorry, site.pp, typo on my part... :) )

brandones: ces-ci is bundled with the other CI servers, and the other CES servers are laptops in the real world... how's the `if` statement in the `init.pp`? Presumably this is a very temporary thing.

brandones:
Ready for re-review, @mogoodrich

PR #80, opened 2020-02-25
Uhm 4234 disk cleanup script

jmbabazi:


mirebalais-modules/scripts/manifests/init.pp
```diff
@@ -0,0 +1,24 @@
+class scripts (
+    $tomcat = hiera('tomcat'),
+    $archive_directory = decrypt(hiera('archive_folder')),
+    $sequence_directory = decrypt(hiera('sequence_folder')),
```
mogoodrich: Unless you are encrypting these two files, just having a "hiera('archive_folder')" and "hiera("sequence_folder")" should be enough, no "decrypt" needed.

manifests/site.pp
```diff
@@ -55,6 +55,9 @@
   include openmrs::backup
   include crashplan
   include mirebalais_reporting::production_setup
+
+  include scripts
```
mogoodrich: Instead of creating a whole new module "scripts" for this, I would suggest just including this in the openmrs module, and adding the jobs to install the cleanup script and set up the cron job to the existing backup.pp, where we set up the mysql archive script.

mirebalais-modules/scripts/templates/cleandiskspace.sh.erb
```diff
@@ -0,0 +1,13 @@
...
+seq_backup_dir='<%= @sequence_directory %>'
+
+disk_size="$(df -k /dev/dm-0 | tail -1 | awk '{print $4}')"
+
+
+if (($disk_size<50000000)); then
+  # find all files under /home/tomcat7/backups/sequences/ and /home/tomcat7/backups/archive/ that are more than 3 days old and delete them
+  find $seq_backup_dir -name "*.7z" -type f -mtime +3 -exec rm -f {} \;
+  find $arch_backup_dir -name "*.7z" -type f -mtime +3 -exec rm -f {} \;
+fi
```
mogoodrich: I'm wondering if instead of creating this whole new script, we could change the existing mysqlarchive.sh to also check for the available disk size and be more aggressive about it's clean-up if disk space is low?

I also think we should have some way to make sure that we never delete *all* the backup zips.  In the "happy path", this should never happen, but what if for some reason the backups fail for 3 or more days and so there are no backups within the past 3 days?

We also might want to just regularly clean up all the backups in the "archives" folder that are more than a year old?

jmbabazi: So, on am... we already have these backups. This is just on prod. The 3 days or more is to ensure that everything is sent to am... Will rethink about the scenario where backup fails to send to a.... for more than 3 days and see if there are adjustments I could add to the script.

Yeah, I thought about the mysql archive approach and now that I'm thinking about it again, Its a good ideacto have this in mysql archive script.

manifests/site.pp
jmbabazi: Thanks @mogoodrich , I thought about this too. My approach here was more about doing some house keeping so that we could make it more easier for other implementer to read
for example, instead of doing
include openmrs::script_name 
we do
include script on the node that we'd like to have the scripts installed.
i.e The implementors would know that there is a module called scripts that contains all the necessary scripts that'll be needed for a specific server 
Note: This is something to think about. I'll change the move the scripts to mysqlarchive,sh

PR #86, opened 2020-03-17
Initial commit for petl module

mseaton:


mirebalais-modules/petl/manifests/init.pp
```diff
@@ -0,0 +1,144 @@
...
+    content => template("petl/application.yml.erb"),
+    owner   => $petl,
+    group   => $petl,
+    mode    => "0755",
+    require => File["/home/$petl/bin/petl.jar"]
+  }
+
+  # Set up scripts and services to execute PETL
+
+  package { "openjdk-8-jdk":
```
mogoodrich: any reason we don't just use the main java mirebalais-module java here?  You can have dependencies between modules.  It's annoyingly  going to run into issues if it runs into two "openjdk-8-jdk" packages (you may have already run into this this morning)

PR #87, opened 2020-03-19
Add puppet parameters for SPA and SPA CI

brandones:
Packages are built by Bamboo in the [MFE project](https://bamboo.pih-emr.org/browse/MFE).

### DIFF
--- a/mirebalais-modules/openmrs/manifests/spa.pp
+++ b/mirebalais-modules/openmrs/manifests/spa.pp
@@ -1,11 +1,24 @@
 class openmrs::spa (
-  $tomcat = hiera('tomcat')
+  $tomcat = hiera('tomcat'),
+  $spa_ci = hiera('spa_ci')
 ) {
+if ($spa_ci) {
+  file { "/home/${tomcat}/.OpenMRS/frontend":
+    ensure => directory
+  }
 
+  file { "/home/${tomcat}/.OpenMRS/frontend/import-map.json":
+    ensure  => file,
+    source  => "puppet:///modules/openmrs/import-map.ci.json",
+    require => [ Package[$tomcat], File["/home/${tomcat}/.OpenMRS/frontend"] ],
+    notify  => Service[$tomcat]
+  }
+} else {
   $package_url = "https://bamboo.pih-emr.org:81/spa-repo/unstable/pih-spa-frontend.zip"
 
   exec{'download_and_install_spa_frontend':
     command => "/usr/bin/wget -q ${package_url} -O /tmp/pih-spa-frontend.zip && unzip -o /tmp/pih-spa-frontend.zip && rm -rf /home/${tomcat}/.OpenMRS/frontend && mv openmrs/frontend/ /home/${tomcat}/.OpenMRS/",
     require => [ Package['unzip'] ]
   }
+}
 }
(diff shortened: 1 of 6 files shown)


brandones:
Tested in virtualbox.

Screenshot from 2020-03-19 16-05-43

mirebalais-modules/openmrs/manifests/spa.pp
```diff
@@ -1,11 +1,23 @@
 class openmrs::spa (
-  $tomcat = hiera('tomcat')
+  $tomcat = hiera('tomcat'),
+  $spa_ci = hiera('spa_ci')
```
mogoodrich: So the idea

mirebalais-modules/openmrs/manifests/spa.pp
```diff
@@ -1,11 +1,23 @@
...
 ) {
+if ($spa_ci) {
```
mogoodrich: So the idea here is if the "spa" parameter is set to true, we will be installing the spa application... and then if spa_ci is set to "true" it installs the custom import map, but otherwise it installs the latest "pih-spa-frontend.zip" artifact that has been build as part of our build pipeline?

What is the advantage of having the custom import map in puppet?  Is this so we can enable and demo features that we aren't even currently considering adding to the PIH-EMR?  

Otherwise, can we handle this by instead of creating a custom import map just having a flag that determines whether to install "pih-spa-frontend.zip" from the stable vs unstable directory?  And then create a "promote to gold" job that copies the latest "pih-spa-frontend.zip" version from the "unstable" directory to the "stable" directory?

mirebalais-modules/openmrs/manifests/spa.pp
brandones: ?

brandones: Oh, found it

mirebalais-modules/openmrs/manifests/spa.pp
brandones: There's sort of three different types of versions we can think about:

- ci: updated whenever any ESM is updated. Always the latest master of every ESM.
- unstable: updated whenever spa-frontend is updated. Contains the deployed versions of ESMs that the latest master of spa-frontend points to.
- stable: doesn't exist yet. But we would manually "promote to gold" from unstable, same as we do with pih-emr.

Does that answer some of these questions?

mirebalais-modules/openmrs/manifests/spa.pp
mogoodrich: Yeah, don't know how that happened.

mirebalais-modules/openmrs/manifests/spa.pp
brandones: Also, if you (or @mseaton ) think that any of these types of versions are superfluous or that there are others we should create, I'm open to making changes to this schema. This is just what I've wound up with in the course of setting things up.

mogoodrich: Yeah, I think that answers my question!  Let me mull it over an we can discuss more next week... trying to think of the balance between the flexibility vs confusion of having three different versions.

PR #93, opened 2020-04-03
Use spa pkg corresponding to config

brandones:
@mogoodrich Just want to run this by you real quick (hence draft), to make sure this isn't a stupid way of implementing this


mirebalais-modules/openmrs/manifests/spa.pp
```diff
@@ -13,10 +14,13 @@
     require => [ Package[$tomcat], File["/home/${tomcat}/.OpenMRS/frontend"] ]
   }
 } else {
-  $package_url = 'https://bamboo.pih-emr.org:81/spa-repo/pih-spa-frontend/unstable/pih-spa-frontend.zip'
+  $config_suffix = regsubst($config_name, 'openmrs-config-', '')
```
mogoodrich, line 17: fwiw, I'd be find with changing the config_name parameter to *only* be the suffix (ie. config_name: ces) so you don't have to go through the rigamarole to retrieve it here.  Probably sets a better precedent as well.

mirebalais-modules/openmrs/manifests/spa.pp
```diff
@@ -13,10 +14,13 @@
...
+  $package_url = "https://bamboo.pih-emr.org:81/spa-repo/pih-spa-frontend/unstable/pih-spa-frontend-${config_suffix}.zip"
```
mogoodrich, line 18: Do we have a plan yet for how we might to version these?  (If the answer is no, and that we don't need to worry about that yet, I'm fine with that for now, because if there was it might affect my advice here)

brandones: No plan yet, but I'm open to worrying about it. I'm also feeling ~~~inspired~~~ by [this howto](https://www.stefanoscerra.it/release-versioning-of-web-application-in-bamboo-build/) to have the versioning respect the node package.json version.

mirebalais-modules/openmrs/manifests/spa.pp
brandones: Yeah, I honestly can't really tell which way is better. What's your reasoning that it would set a better precedent?

mogoodrich: I was thinking that we needed to add a third dependency of some sort it would clearly set the precedent that the naming convention should be:

some-other-thingy-{config_suffix}.zip

Seems like it would be less likely that people would miss this convention (though they still could) and we would end up with something like this down the road:

openmrs-config-ces.zip
pih-spa-frontend-ces.zip
some-other-thingy-**mexico**.zip

brandones: Yeah, that makes sense to me.

brandones:
@mogoodrich This is working and ready for review now, if you wanna take another look real quick

mirebalais-modules/openmrs/manifests/init.pp
```diff
@@ -11,8 +11,7 @@
     $tomcat = hiera('tomcat'),
     $tomcat_webapp_dir = hiera('tomcat_webapp_dir'),
     $junit_username = hiera('junit_username'),
-    $junit_password = decrypt(hiera('junit_password')),
-    $pih_config = hiera('pih_config'),
+    $junit_password = decrypt(hiera('junit_password')), $pih_config = hiera('pih_config'),
```
mogoodrich, line 14: Any reason this ended up on one line?  autoformatting?

mirebalais-modules/openmrs/manifests/init.pp
```diff
@@ -195,7 +194,7 @@
     }
 
     exec{'install-openmrs-configuration':
-      command => "rm -rf /tmp/configuration && unzip -o /tmp/${config_name}.zip -d /tmp/configuration && rm -rf /home/${tomcat}/.OpenMRS/configuration && mkdir /home/${tomcat}/.OpenMRS/configuration && mv /tmp/configuration/* /home/${tomcat}/.OpenMRS/configuration",
+      command => "rm -rf /tmp/configuration && unzip -o /tmp/${config_name}.zip -d /tmp/configuration && rm -rf /home/${tomcat}/.OpenMRS/configuration && mkdir /home/${tomcat}/.OpenMRS/configuration && cp -r /tmp/configuration/* /home/${tomcat}/.OpenMRS/configuration",
```
mogoodrich, line 197: I noticed this line failing on one of the builds yesterday, but when I tried it manually it worked and then when I ran it again it passed... were you running into a similar problem?  (Just curious, nothing wrong with this chagne)

mirebalais-modules/openmrs/manifests/spa.pp
```diff
@@ -1,22 +1,42 @@
...
 ) {
+  # Please see
+  # https://pihemr.atlassian.net/wiki/spaces/PIHEMR/pages/555188230/CI+Release+and+Deployment
+  # for more information about this process
 if ($spa_ci) {
-  file { "/home/${tomcat}/.OpenMRS/frontend":
-    ensure => directory
+  # In CI, the import map is located on the server, but refers to modules which
+  # are served by Apache on the Bamboo server. The site-specific assets also
+  # are copied onto the server.
```
mogoodrich, line 12: @brandones forgive me because I know I've asked this before, but I always get confused... :) 

In the "deployment" release all imported artifiacts get bundles into the frontend zip and deployed, but in the CI instance they are pulled directly from bamboo, correct?  This makes sense *but* if we are setting up the build now to build a new spa-frontnend.zip for each country upon every commit, is there really any difference between these two?

mirebalais-modules/openmrs/manifests/init.pp
brandones: Whoops! Accidental

mirebalais-modules/openmrs/manifests/init.pp
brandones: I imagine that was while I was debugging against ces-ci (before I started debugging in VirtualBox, which was the right thing to do, and much faster)? Lots of wonky things were happening, and expected failures.

If it was on some other box, that might be cause for concern.

mirebalais-modules/openmrs/manifests/spa.pp
brandones: Correct. Yes, the difference is this:

spa-frontend.zip tracks the latest commits to both pih/spa-frontend and the country config repo

CI tracks the latest commits of the above *plus* every SPA module

mogoodrich: Got it, so when we do want to upgrade the "prod" release with the latest version of a SPA module (other than spa-frontend) where do we do that?

brandones: Install that SPA module to [spa-frontend/frontend-package](https://github.com/PIH/spa-frontend#packagesfrontend-package)

PR #95, opened 2020-04-10
CI servers should refer to import-map on bamboo 

brandones:
What remains to do is updating the Bamboo jobs so that 
- ESM build jobs update the import maps on Bamboo/Apache and
- Config repo build jobs update the config.json file on Bamboo/Apache


mogoodrich:
LGTM!

PR #96, opened 2020-04-13
Add script for Microfrontends CI

brandones:


brandones:
Whoops, forgot to include support for overrides. We at least need to override @openmrs/esm-root-config .

brandones:
Fixed. Ready to go.

scripts/build_mfe_and_update_ci_import_maps.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+
+set -o xtrace  # print each command
+set -e         # die on error
+
+#=== Get package version number
+app_version="$(node -e 'console.log(require("./package.json").version);')"
+version="${app_version}.${bamboo_buildNumber}"
+
+#=== Copy to build dir
```
mogoodrich, line 10: Minor point, but might be good to add something like "spa-repo on Bamboo".... because this is really a distribution directory as much as a build directory, right?  I didn't understand exactly what the command meant until I dug into the code.

brandones: Whoops, my comment was just completely misleading -- it's a distribution dir, not a build dir at all. Thanks.

PR #98, opened 2020-04-29
azcopy and percona and archive cleanup

jmbabazi:


mirebalais-modules/openmrs/manifests/backup.pp
```diff
@@ -43,6 +45,39 @@
...
+    mode    => '0700',
+    owner   => 'root',
+    group   => 'root',
+    content => template('openmrs/backupAzure.sh.erb'),
+    require => File["/usr/local/sbin/azcopy"]
+  }
+
+  cron { 'backup-az':
+    ensure  => present,
+    command => 'sh /usr/local/sbin/backupAzure.sh &> /dev/null',
```
brandones, line 73: `bash` not `sh`, right? And do you really want to be losing stderr, rather than letting it log normally?

mirebalais-modules/openmrs/templates/backupAzure.sh.erb
```diff
@@ -0,0 +1,28 @@
+#/bin/bash
```
brandones, line 1:
```suggestion
#!/bin/bash
```

mirebalais-modules/openmrs/templates/backupAzure.sh.erb
```diff
@@ -0,0 +1,28 @@
...
+bold=$(tput bold)
+normal=$(tput sgr0)
+red="\033[0;31m"
+
+mkdir -p /home/tomcat7/backups/logs
+log_file=/home/tomcat7/backups/logs/azcopy.logs
+
+echo -e "\e[32m"${bold}$(date +%F-time-%T)${normal} >> $log_file
+
+rm -rf $log_file
```
brandones, line 20: This adds content to the log file, only to delete it immediately?

mirebalais-modules/openmrs/templates/backupAzure.sh.erb
```diff
@@ -0,0 +1,28 @@
...
+azcopy copy "/home/tomcat7/backups/sequences/${BACKUPFILE}"  "${AZURL}/${AZPATH}/sequence/${BACKUPFILE}?sv=2019-02-02&ss=bfqt&srt=sco&sp=rwdlacup&se=2025-03-29T22:00:00Z&st=2020-03-30T13:00:00Z&spr=https&sig=${AZSECRET}" --recursive=true >> $log_file
+
+#Remove a month old logs
+if(($DAY == 02));
+then
+        rm -rf root/.azcopy/*.log
+        rm -rf root/.azcopy/plans/*
+fi
```
brandones, line 28: This deletes all the log files and everything in plans/ once a month, not just the "month old" stuff. You probably want a variation on the command found [here](https://unix.stackexchange.com/a/205539/306723). And then you can get rid of the `if` block and just run it every time.

brandones:
Mostly reviewed for syntax; I don't really know what this stuff is supposed to do.

mirebalais-modules/openmrs/templates/backupAzure.sh.erb
jmbabazi: Thanks @brandones , when it is the second day of the month every log should be deleted, The idea is to not have logs fill the disk space. There should be every once in a while (preferably at least once a week) manual  checks on the blobs to make sure that there is backups. So if everything word the way it is supposed to.we don't need the logs thus the idea of having an automated deletion on the second day of the month.

mirebalais-modules/openmrs/templates/backupAzure.sh.erb
jmbabazi: Good catch

mirebalais-modules/openmrs/manifests/backup.pp
jmbabazi: Got the idea from https://github.com/Azure/azure-storage-azcopy/issues/413, If you don't specify the those, the copy doesn't work. I've tested with sh, I think bash would work too but I didn't test.

mirebalais-modules/openmrs/templates/backupAzure.sh.erb
```diff
@@ -0,0 +1,28 @@
+#/bin/bash
+export PATH=/usr/local/sbin:/sbin:/bin:/usr/sbin:/usr/bin:/root/bin
+
+cd /home/tomcat7/backups/sequences
+
+BACKUPFILE=$(ls -t | head -n1)
+AZSECRET='<%= @az_secret %>'
+AZURL='<%= @az_url %>'
+AZPATH='haiti/mirebalais'
```
mogoodrich, line 9: @jmbabazi  I think you could use the "site_domain" yaml variable here so that each server puts it in a subdirectory names after the site, ie AZPATH='<%= site_domain %>'....

mogoodrich: Or you could use the backup_file_prefix yam variable?

mirebalais-modules/openmrs/manifests/backup.pp
```diff
@@ -43,6 +45,39 @@
...
+    user    => 'root',
+    hour    => '*/7',
+    minute  => 20,
```
mogoodrich, line 76: I'm forgetting my cron format... when does this execute?

mirebalais-modules/openmrs/templates/backupAzure.sh.erb
```diff
@@ -0,0 +1,28 @@
...
+normal=$(tput sgr0)
+red="\033[0;31m"
+
+mkdir -p /home/tomcat7/backups/logs
+log_file=/home/tomcat7/backups/logs/azcopy.logs
+
+echo -e "\e[32m"${bold}$(date +%F-time-%T)${normal} >> $log_file
+
+rm -rf $log_file
+azcopy copy "/home/tomcat7/backups/sequences/${BACKUPFILE}"  "${AZURL}/${AZPATH}/sequence/${BACKUPFILE}?sv=2019-02-02&ss=bfqt&srt=sco&sp=rwdlacup&se=2025-03-29T22:00:00Z&st=2020-03-30T13:00:00Z&spr=https&sig=${AZSECRET}" --recursive=true >> $log_file
```
mogoodrich, line 21: I'm not familar with the parameters here (sv, ss, srt, sp, se, st... ) so I can't review them in much detail.... what are they for?  I see that some are hardcoded to certain date?

mirebalais-modules/openmrs/templates/backupAzure.sh.erb
mogoodrich: Defer to you @jmbabazi about when you delete the logs, but I do agree that the comment is bit confusing and implies it's only deleting logs a month or old... could be "#Remove all logs on the 2nd of each month".

mirebalais-modules/openmrs/templates/backupAzure.sh.erb
```diff
@@ -0,0 +1,28 @@
+#/bin/bash
+export PATH=/usr/local/sbin:/sbin:/bin:/usr/sbin:/usr/bin:/root/bin
+
+cd /home/tomcat7/backups/sequences
+
+BACKUPFILE=$(ls -t | head -n1)
```
mogoodrich, line 6: So this copies just the most recent backup each time it runs?  That means that if it doesn't run for a day or two, then backups from those days will be lost?

I really think we want to consider using the "azcopy sync" parameter to sync the entire directory.  It makes the code a little simpler as well.  Was there a reason not to do this I'm missing:

https://docs.microsoft.com/en-us/azure/storage/common/storage-ref-azcopy-sync

mirebalais-modules/openmrs/templates/mysqlarchive.sh.erb
```diff
@@ -29,18 +29,4 @@ then
...
-
-# Delete folders in the archive directory only if the disk space is greater than 60 GB
-disk_size="$(df -k /dev/dm-0 | tail -1 | awk '{print $4}')"
-#
-if (($disk_size<60000000)); then
-# find all files under /home/tomcat7/backups/sequences/ and /home/tomcat7/backups/archive/ that are more than a day old and delete them
- find ${DIR}/archive/${YEAR} -name "*.7z" -type f -mtime +15 -exec rm -f {} \; 2>/dev/null
-fi
\ No newline at end of file
+find ${DIR}/sequences/ -name "*.7z" -type f -mtime +30 -exec rm -f {} \;
```
mogoodrich, line 32: I'm happy to simplify this but did we confirm that we no longer need all the other stuff below?  I though we were deleting more frequently on Mirebalais because of disk space on that machine?

mirebalais-modules/openmrs/templates/backupAzure.sh.erb
brandones: Okay, but what if something goes terribly wrong on the first day of the month, and you find out about it on the second or third day of the month? Don't you always care about the same number of days back in the logs, whether it's the first or the fifteenth? So maybe if a month of logs is more than we want to store, we could always store the last week, two weeks, three days, whatever?

brandones: Just pitching these ideas in -- as with @mogoodrich , I defer to your decision, @jmbabazi .

mogoodrich: I agree with @brandones  that in general keeping logs back a month would be better, but I don't really know how helpful the azcopy logs are or what they show.

mseaton: This is the kind of thing that logrotate can help manage, which I think we already use for other logging on the servers...

jmbabazi: This is how a log file would look like. Its all about the network error.

`2020/04/29 16:55:40 ==> REQUEST/RESPONSE (Try=1/639.711826ms, OpTime=38.761134813s) -- RESPONSE SUCCESSFULLY RECEIVED
   PUT https://XX/backups/haiti/mirebalais/sequence/mirebalais_backup_20200429-
   X-Ms-Request-Id:

2020/04/29 16:55:40 ==> REQUEST/RESPONSE (Try=1/405.754709ms, OpTime=39.03831149s) -- RESPONSE SUCCESSFULLY RECEIVED
   PUT https://XXX/haiti/mirebalais/sequence/mirebalais_backup_20200429-013001.sql.7z?blockid=

The azcopy.log should have the logs of errors caused by the script and the status of the copied file itself. 

Its okay to remove those file after a month because some of the sizes of the cloud machines or at some sites are small. And as I had mentioned regular blob checks is required.

mirebalais-modules/openmrs/templates/backupAzure.sh.erb
jmbabazi: Thats for easy reading of the logs files, The date and time when the script was executed.

mirebalais-modules/openmrs/templates/backupAzure.sh.erb
jmbabazi: I'm not sure about using sync for sites with low bandwidths. Also, I'm not sure what would happen in the blob when the 30 days old backups are deleted by the mysqlarchive script from the host machine. Will this be deleted from the blob because sync mirrors what is in the sequence folder? This is why I used the copy. The copy sends over the latest file got from `BACKUPFILE=$(ls -t | head -n1)` and this file will be permanent in the blob

mirebalais-modules/openmrs/templates/mysqlarchive.sh.erb
jmbabazi: For the stuff below, I don't think we need them. They were meant for mirebalais only, but I realized that even on other sites they are deployed. The last thing we should do is to delete backups from machines that have enough space. This is why I changed the file back to its original version.

jmbabazi: I'm thinking of introducing a new file specific to mirebalais or doing this via the cron tabs.

mirebalais-modules/openmrs/templates/backupAzure.sh.erb
jmbabazi: `So this copies just the most recent backup each time it runs? That means that if it doesn't run for a day or two, then backups from those days will be lost?
`
The cron is setup to run everyday and at least twice in a day. I'm not sure what would cause the cron not to run for servers that always up. If you think that there is any other reason that would cause this not to run, please let me know

mirebalais-modules/openmrs/manifests/backup.pp
jmbabazi: This executes every 7 hours.

mirebalais-modules/openmrs/templates/backupAzure.sh.erb
mogoodrich: Yes, I just meant if the server was down during the the time the cron ran, or if the Internet connection was down and so the update failed.

It's a good question about deleting the old backups if we use "sync"... I think there's a parameter not ever delete files, but, if there isn't, I agree it isn't a great solution.

mirebalais-modules/openmrs/templates/mysqlarchive.sh.erb
mogoodrich: Makes sense, thanks!

mirebalais-modules/openmrs/templates/backupAzure.sh.erb
mogoodrich: But should they not be hardcoded to certain dates then?

jmbabazi: So I've given another thought on this, instead of removing the azcopy.log file every time the script executes, I am going to append the outputs to this one file. 

I'm not sure I understood your last question but my idea to to have all the logs and their dates within one file:- not multiple files that are created each time the script  executes.

mogoodrich: @jmbabazi  I meant the dates in the azcopy copy line, like /${BACKUPFILE}?sv=**2019-02-02** ... but looking into I see these are just part of the SAS token generated for you, so never mind, looks good!

PR #99, opened 2020-05-13
Use mirebalais-latest SPA packages for sites with /mirebalais base URL

brandones:
See https://github.com/PIH/spa-frontend/pull/6


mogoodrich:
Oh, annoying... so we end up generating two identical zip packages for ZL except for the prefix?

PR #100, opened 2020-05-21
Update all CES servers to Release 96

brandones:
https://pihemr.atlassian.net/browse/UHM-4576


brandones:
Oh yeah! Good point. I'll make that change with Alfred next week. Thanks @mogoodrich .

PR #103, opened 2020-07-20
removing the old scripts for copying the HUM Prod database from prod …

mogoodrich:
…to reporting


mogoodrich:
@jmbabazi I noticed this stuff while working on the PETL stuff.  I think we can just remove all this at this point?  Or should we keep it around for a bit longer? Thanks!

brandones (approved):
Would be great if @jmbabazi could take a look

brandones:
Ping @jmbabazi

PR #105, opened 2020-08-06
Change ces-capitan and ces-laguna back to configuring the laptop-servers.

brandones:
The cloud EMRs should have their own files like ces-capitan.pih-emr.org.yaml


brandones:
In collaboration with @Alfredreyescruz , who is presently fixing those servers.

PR #107, opened 2020-09-02
Uhm 4668 letsencrypt

jmbabazi:


hieradata/ami.pih-emr.org.yaml
```diff
@@ -9,6 +9,12 @@ mysql_root_password: ENC[oWqIfmn/ER4CYIvjJSpIzQ==]
 mysql_bind_address: 0.0.0.0
 openmrs_db_password: ENC[BNK3oX/1a+ta7Lal9jVd1A==]
 
+#letsencrypt
+azure_dns_subscription_id: ENC[roYSeZU0cSjB6+P6aOWEXYVgCL64D33EQcfUd7L80970icq7M7LFmvlOSyYPI3bm]
+azure_dns_tenant_id: ENC[5LWTjj76sYS2kxC4FktrecjxHpznSDVezymGgwlCzdNGmdymse5ihQOOKFpAwCJa]
+azure_dns_app_id: ENC[shiW8Zw+zKNZjwEojjsm0pzj3SOJGUpRJl7pMZ9FULbrxVKYi/nWrAmRAGifWya3]
+azure_dns_client_secret: ENC[f+Bvtjd7pyxAFpRgF3ZM8kzj5BfsQY3bZJ3jZrlebuPs5yF4zR69KWwP8l98Ut+5]
```
mogoodrich, line 16: Are any of these shared among all instances and could be moved to the common.yml?

mirebalais-modules/apache2/templates/install-letsencrypt.sh.erb
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+set -x
+
+EMRREPORTING='<%= @site_domain %>'
```
mogoodrich, line 5: Shouldn't this variable be called SITE_DOMAIN since it is no longer specific to the EMRREPORTING server?

mirebalais-modules/apache2/templates/install-letsencrypt.sh.erb
```diff
@@ -0,0 +1,28 @@
...
+./<%= @letsencrypt_user %>.sh --install
+
+chown -R  <%= @letsencrypt_user %>:<%= @letsencrypt_user %> /var/<%= @letsencrypt_user %>/.<%= @letsencrypt_user %>.sh
+
+cd /var/<%= @letsencrypt_user %>
+
+export AZUREDNS_SUBSCRIPTIONID='<%= @azure_dns_subscription_id %>'
+export AZUREDNS_TENANTID='<%= @azure_dns_tenant_id %>'
+export AZUREDNS_APPID='<%= @azure_dns_app_id %>'
+export AZUREDNS_CLIENTSECRET='<%= @azure_dns_client_secret %>'
```
mogoodrich, line 24: How are these variables used?  Just trying to understand the process beter.

mirebalais-modules/apache2/manifests/init.pp
```diff
@@ -93,88 +97,96 @@
...
+  }
+
+  exec { "add user to sudoers":
+    unless  => "/bin/cat /etc/sudoers | grep $letsencrypt_user",
+    command => "echo '$letsencrypt_user    ALL=(ALL) NOPASSWD: /usr/sbin/service apache2 restart' | sudo EDITOR='tee -a' visudo",
+    require => User["$letsencrypt_user"]
+  }
+
+  exec { "download and from the git repo":
+    command => "rm -rf /var/$letsencrypt_user/acme.sh && git clone https://github.com/acmesh-official/acme.sh.git /var/$letsencrypt_user/acme.sh",
```
mogoodrich, line 147: I also think we should just hardcode "acme" instead of using $letsencrypt_user... "Acme" is the application name, correct, so regardless of the user that runs it, I would think it would live in a "/var/acme" directory structure.

We also should be able to do a wget to fetch this instead of a full git clone.

mirebalais-modules/apache2/manifests/init.pp
```diff
@@ -93,88 +97,96 @@
...
-      ensure => file,
-      source => "puppet:///modules/apache2/etc/ssl/certs/${ssl_chain_file}",
-      notify => Service['apache2']
-    }
-
-    file { "${ssl_key_dir}/${ssl_key_file}":
-      ensure => present,
-      notify => Service['apache2']
-    }
+  user { "$letsencrypt_user":
```
mogoodrich, line 100: I'm thinking we can just hardcode this to "acme".  Or, if you said, if we think we can just have root execute this, not create this user at all.

mirebalais-modules/apache2/manifests/init.pp
```diff
@@ -93,88 +97,96 @@
...
+    ensure => 'present',
+    password => "$letsencrypt_user_password",
```
mogoodrich, line 102: Since we will never actually need to log in as this user, do they need a password? When we create the Tomcat user I don't think we assign them a password?

mirebalais-modules/apache2/templates/install-letsencrypt.sh.erb
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+set -x
+
+EMRREPORTING='<%= @site_domain %>'
+
+export HOME=/var/<%= @letsencrypt_user %>
```
mogoodrich, line 7: Continuing with the previous comments, I'm hoping we could just made this "/var/acme"

mirebalais-modules/apache2/manifests/init.pp
```diff
@@ -93,88 +97,96 @@
...
+  # the unless condition only allow this script to run once.
+  exec { "download and run install letsencrypt using $letsencrypt_user":
+    unless  => "/bin/ls -ap /var/$letsencrypt_user | grep '^\..*/$' | grep acme.sh | grep -v grep",
+    command => "/var/$letsencrypt_user/install-letsencrypt.sh",
+    require => [User["$letsencrypt_user"], Exec['download and from the git repo']]
+  }
+
+  cron { "renew certificates using $letsencrypt_user user":
+    ensure  => present,
+    command => "'/var/$letsencrypt_user/.$letsencrypt_user.sh'/$letsencrypt_user.sh --cron --home '/var/$letsencrypt_user/.$letsencrypt_user.sh' > /dev/null",
```
mogoodrich, line 160: Here, too, I think we can just hardcode to "acme" and "acme.sh".

The nested ".acme.sh/acme.sh also confuses me, but I see that's the way they do it in the Acme examples, so I can live it with it! :)

mirebalais-modules/apache2/templates/install-letsencrypt.sh.erb
```diff
@@ -0,0 +1,28 @@
...
+chown -R  <%= @letsencrypt_user %>:<%= @letsencrypt_user %> /var/<%= @letsencrypt_user %>/.<%= @letsencrypt_user %>.sh
+
+cd /var/<%= @letsencrypt_user %>
+
+export AZUREDNS_SUBSCRIPTIONID='<%= @azure_dns_subscription_id %>'
+export AZUREDNS_TENANTID='<%= @azure_dns_tenant_id %>'
+export AZUREDNS_APPID='<%= @azure_dns_app_id %>'
+export AZUREDNS_CLIENTSECRET='<%= @azure_dns_client_secret %>'
+
+.<%= @letsencrypt_user %>.sh/<%= @letsencrypt_user %>.sh --issue --dns dns_azure -d "$EMRREPORTING" --force
```
mogoodrich, line 26: this could just be .acme.sh/acme.sh...

mirebalais-modules/apache2/templates/install-letsencrypt.sh.erb
jmbabazi: Good catch, I''l change that.

mirebalais-modules/apache2/templates/install-letsencrypt.sh.erb
jmbabazi: These are api keys. the acme script uses the script to do a verification with the dns record https://github.com/acmesh-official/acme.sh/wiki/dnsapi

mirebalais-modules/apache2/manifests/init.pp
jmbabazi: Ok, I think I like the idea of not having this user

mirebalais-modules/apache2/templates/install-letsencrypt.sh.erb
jmbabazi: I pinged you on why I had decided to go with this approach but if we are going to remove this user, then I agree that we don;t need these encrypted.

PR #110, opened 2020-09-04
Update README.md

brandones:


mogoodrich (approved):
LGTM, but @jmbabazi  can you confirm this is accurate?

jmbabazi:
Thanks @brandones. Quick question, did you have any issues running or install letsencrypt, if yes let me know. I think its worth solving the issue in puppet.

brandones:
@jmbabazi  It didn't work the first time because I didn't have `site_domain` set. Then it didn't re-run, so I did `rm -r /var/acme` and ran `puppet-apply` again. It works generally, I just didn't have things set up right :)

jmbabazi:
Thanks @brandones, This is what causes the above not to re-run again https://github.com/PIH/mirebalais-puppet/blob/6e49424bc243c866ed36ed48b80e64d0671016f8/mirebalais-modules/apache2/manifests/init.pp#L146 . The reason for this is because when you request/re-install the certs many times and it exceed the limits. blank certs are created. Thanks for adding the ReadMe in and feel free to delete this branch.

PR #116, opened 2020-09-05
Adding azcopy vars to common.yaml

jmbabazi:


jmbabazi:
@mogoodrich, would you please review these. I changed the localhost to 127.0.0.1 because openmrs webapp was inaccessible while using localhost on jjdossen

mogoodrich (approved):
LGTM, thanks @jmbabazi !

PR #117, opened 2020-09-09
Lib 02

BranfordTGbieor:
fyi @jmbabazi  @mogoodrich

### DIFF
--- a/hieradata/jjdossen.pih-emr.org.yaml
+++ b/hieradata/jjdossen.pih-emr.org.yaml
@@ -2,8 +2,6 @@
 site_domain: jjdossen.pih-emr.org
 
 az_backup_folder_path: liberia/jjdossen
-az_secret: ENC[1oQfg4Gi51CDU3XTOIkHatJn2W+vAxx1Juj/wYnVwyRtEFk7hm7XAkA0CbKlsUcUdED4rBiSDIYAlTiO3aBryg==]
-az_url: ENC[6NweILo1Eq7pKNo88HD1o/fyAWOijtE/7o3Pn+pd8tJ9q2Z1PerwvhtXMsVi2nl37EjQL96k8gApwQCE+nKnaw==]
 
 package_release: 'stable/'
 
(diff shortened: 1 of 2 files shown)


jmbabazi (approved):
Thanks @BranfordTGbieor , LGTM. fyi @mogoodrich

hieradata/jjdossen.pih-emr.org.yaml
```diff
@@ -1,6 +1,10 @@
 ---
 site_domain: jjdossen.pih-emr.org
 
+az_backup_folder_path: liberia/jjdossen
+az_secret: ENC[1oQfg4Gi51CDU3XTOIkHatJn2W+vAxx1Juj/wYnVwyRtEFk7hm7XAkA0CbKlsUcUdED4rBiSDIYAlTiO3aBryg==]
+az_url: ENC[6NweILo1Eq7pKNo88HD1o/fyAWOijtE/7o3Pn+pd8tJ9q2Z1PerwvhtXMsVi2nl37EjQL96k8gApwQCE+nKnaw==]
```
mogoodrich, line 6: @jmbabazi  I'm forgetting, does the az server and url differ from server to server, or can it just be inherited from "common.yaml"?

jmbabazi: Its the same, we added it in the common.yaml recently.

mogoodrich: Thanks @jmbabazi ... so then @BranfordTGbieor  we don't need to add "az_secret" and "az_url"  here (or in pleebo), we can just add the "az_backup_folder_path".  Pleebo and JJ Dossen will inherit the secret and the url from common.yaml.

BranfordTGbieor: OK! Its the same with the letsencrypt certs properties. We removed them because they inherit from the common.yaml file too.

Thanks @mogoodrich.

BranfordTGbieor (commented):
@

hieradata/jjdossen.pih-emr.org.yaml
BranfordTGbieor: OK! Its the same with the letsencrypt certs properties. We removed them because they inherit from the common.yaml file too.

Thanks @mogoodrich

PR #118, opened 2020-09-18
petl logrotate, ntp fixes, fixes to issue identified by cron

jmbabazi:


jmbabazi:
@mogoodrich , please find the PR. This is what was done.
* introduced resolvconf to facilitate DNS resolutions
* I removed the local ntp servers on some yaml files.
* removed remote_db_server on some yaml files that use azcopy
* fix for cron issue "service apache2 restart"

mirebalais-modules/ntpdate/manifests/init.pp
```diff
@@ -1,7 +1,15 @@
 class ntpdate(
   $timezone = hiera('server_timezone'),
+  $server_1 = hiera('ntp_server_1'),
+  $server_2 = hiera('ntp_server_2'),
+  $server_3 = hiera('ntp_server_3'),
+  $server_4 = hiera('ntp_server_4')
```
mogoodrich, line 6: Strange!  I don't remember why i removed these previously!  Thanks!

hieradata/emr.hum.ht.yaml
```diff
@@ -8,13 +8,9 @@ config_name: openmrs-config-zl
 config_version: 1.4.0
 
 server_timezone: America/New_York
-ntp_server_1: 192.168.1.151
-ntp_server_2: 192.168.1.174
-ntp_server_3: 192.168.1.151
-ntp_server_4: 192.168.1.174
```
mogoodrich, line 14: There was a reason we added this back in the day... I don't 100% remember why, but it looks like they were adding in March 2016 so I'm guessing it was because of the weirdness involving daylight savings in 2016...  I'm fine with removing and we can monitor and then add again if needed.

mogoodrich (approved):
LGTM, thanks @jmbabazi !  Let's just review the log rotate works correctly on HUM-CI after it rolls out, and make sure we check the ntp service on the Mirebalais machines after the next update.

PR #119, opened 2020-09-29
Improvement to scripts

jmbabazi:
@mogoodrich , I made some improvements to the archive and azcopy scripts. Would you please review.


mogoodrich (commented):
@jmbabazi  I haven't been following how azcopy works 100%, but generally looks good to me! If you have any particular questions for me, lmk...

PR #121, opened 2020-10-20
Install Petl error script

jmbabazi:


hieradata/common.yaml
```diff
@@ -115,6 +115,8 @@ petl_sqlserver_port: 1433
 petl_sqlserver_databaseName: openmrs_extractions
 petl_sqlserver_user: foo
 petl_sqlserver_password: bar
+petl_error_subject: Petl errors
+petl_cron_hour: '*/6'
```
mogoodrich, line 119: Not a bit deal, but is there a reason the default cron hour is different from the time we actually use for Haiti HIV Test and HUM-CI?

mogoodrich (approved):
One small comment, but generally LGTM @jmbabazi , thanks!

hieradata/common.yaml
jmbabazi: The common.yaml cron is for productions purposes like the reporting server. I thought we don't want to run it as often since the petl job executes only once a day. Ideally this cron should run once but I made it every six hours as a reminder in cases where someone would forget.

PR #126, opened 2021-02-16
Change all the CES servers to follow the SNAPSHOT config version

brandones:


brandones:
This was done live with Jose Ramon. Merging.

mogoodrich:
We generally don't advise having prod servers pick up snapshots... this means that the different laptops could end up with indeterminate different configurations depending on the time they are updated?

PR #128, opened 2021-03-18
Puppet ubuntu 20.04

mogoodrich:


mirebalais-modules/java/manifests/init.pp
```diff
@@ -15,7 +15,7 @@
   package { 'openjdk-8-jdk':
     ensure  => present,
     notify => Service[$tomcat],
-    require => [File['/etc/environment'], Apt::Ppa['ppa:openjdk-r/ppa']]
```
mogoodrich, line 18: I assume this was required because the openjdk PPA was not available but default in Ubuntu 16.04.  Seems fine to remove it, though I suspect this means that things will fail if we ever try to install a new instance again on Ubuntu 16.04 (but I don't know why we'd ever be doing that.. :) )

mogoodrich (commented):
This generally looks good to me @jmbabazi , though let's discuss/test a little more before merging.  I'm also curious about testing on ci.pih-emr.org, since you've done the 20.04 update there.

mogoodrich (commented):
This generally looks good to me @jmbabazi ... let's review some more next week and test out of the state of the CI server and then we can consider merging this in.

gem-update.sh
```diff
@@ -1,4 +1,41 @@
-#! /bin/bash
+#!/bin/bash
+
+osPackagesToDowload=(
+http://archive.ubuntu.com/ubuntu/pool/main/s/sysvinit/initscripts_2.88dsf-59.3ubuntu2_amd64.deb
+http://archive.ubuntu.com/ubuntu/pool/main/s/sysvinit/sysv-rc_2.88dsf-59.3ubuntu2_all.deb
+http://archive.ubuntu.com/ubuntu/pool/main/i/insserv/insserv_1.14.0-5ubuntu3_amd64.deb
```
mogoodrich, line 6: Just out of curiosity, why do we need to manually install these packages?

gem-update.sh
```diff
@@ -33,5 +70,76 @@ rm -f /var/lib/gems/2.3.0/gems/facter-2.5.7/lib/facter/ec2.rb
...
+	apt-get -y dist-upgrade
+	
+	/usr/bin/apt install -y rake ruby-did-you-mean libruby2.5 ruby2.5
+
+else
+	echo "ruby2.5 is already installed"
+fi
+
+### if mysql-server-5.6 does not exist, install it. If it exists, do not remove
+if ! isPackageInstalled mysql-server-5.6 ; then
```
mogoodrich, line 116: Does this actually install mysql-server-5.6, or just remove versions of mysql that might come pre-installed on Ubuntu 20.04. 

It seems like it does latter, because I only see "apt-get remove" and "apt-get purge" here?  I think this is correct, because Puppet installs MySQL in "mysql_setup", right?  But we should update the comment in this case to say something like "remove any default version of mysql installed if mysql 5.6 has not been installed"?

mogoodrich (commented):
Generally looks good, thanks @jmbabazi ... a couple comments in the code.

Just to confirm, after merging this in, the overall workflow for upgrading each server would be:

* Run the 20.04 upgrade
* Run gem-update.sh, which will update the version of Ruby and Puppet?

Also, unfortunately, it looks like we may still need the hack to remove the "ec2" fact for factor, so we may want to try adding these two lines back into the 20.04 block (the ones I told you remove before, sorry... :) )  We can add these back in post-merge if it's easier because you have everything fully tested:

https://github.com/PIH/mirebalais-puppet/blob/master/gem-update.sh#L30

gem-update.sh
jmbabazi: Thanks @mogoodrich . These startup scripts are required by mysql 5.6. sysrv can be installed directly from the official repo but the initscript package isn't available. The sysvrc package from the official repo is not compatible with the above initscript.

The above is for a use case of installing mysql 5.6 on a new ubuntu 20.04 system. If there is a need to have a vanilla Ubuntu 20.04 system, this will work but hopefully we will move to mysql 8 soon and we won't have to have these workarounds.

gem-update.sh
jmbabazi: This removes the packages of mysql (mysql 5.8) and its dependencies that come pre-installed on Ubuntu 20. Puppet takes care of installing mysql 5.6. I will add the comments

mogoodrich: Cool, that's what I thought, and that sounds perfect, let's just update the comments!

gem-update.sh
mogoodrich: Cool, sounds good @jmbabazi ... let's just add some comments so we remember the reason we are doing this...

mogoodrich (commented):
beyond adding a few comments LGTM @jmbabazi ... thanks for working through all of this!

PR #130, opened 2021-04-08
UHM-5322 - remove owa installation from puppet, in order for this to …

mseaton:
…come from the distribution artifact directly.

I have not tested this at all, but this seems like what we would need to do to remove the owa installation from puppet...


mogoodrich (approved):
LGTM, let's review with James when to merge this all in

PR #133, opened 2021-05-03
Removing unused liquibase steps

jmbabazi:


mogoodrich:
This generally looks good, but I suspect we are going to run into whatever problems I ran into last time with removing (and that I've forgotten the details of!).  Do we want to coordinate when we want to merge this in to a time when we are sure we can test?

brandones:
Any reason not to just delete the code? It will always be available in the git history.

jmbabazi:
Thanks @mogoodrich , lets coordinate when we can merge this in. We can remove this if we think the history is enough @brandones

brandones:
@jmbabazi @mogoodrich Should we close this as stale? Or should we fix it up and merge it in?

mogoodrich:
Probably makes sense to close as stale at this point, thoughts @jmbabazi ?

PR #135, opened 2021-05-12
Add hack to fix webapp name in SPA application 

brandones:
https://pihemr.atlassian.net/browse/UHM-5276

Unfortunately, since SPA path and API path are set at application build time, and since we only have one distribution which we deploy to different servers with different webapp names, we need an ugly hack. This changes the SPA path and API path on servers for which they aren't the default.

### DIFF
--- a/mirebalais-modules/openmrs/manifests/init.pp
+++ b/mirebalais-modules/openmrs/manifests/init.pp
@@ -204,6 +204,13 @@ class openmrs (
     notify => [ Exec['tomcat-restart'] ]
   }
 
+  # hack to change webapp name in the frontend application after it has already been built into the deb
+  exec {
+    unless  => "test ${webapp_name} = openmrs",
+    command => "sed -i 's/\/openmrs\([\/\"]\)/\/${webapp_name}\1/g' /home/${tomcat}/.OpenMRS/frontend/index.html",
+    require => Package['pihemr']
+  }
+
   exec { 'tomcat-restart':
     command     => "service ${tomcat} restart",
     user        => 'root',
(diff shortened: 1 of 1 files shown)


mseaton (approved):
Fine as a short-term hack I guess.  @mogoodrich what's the lift to standardize on "openmrs" and get rid of "mirebalais" altogether?  @brandones any chance that MFE can support this being more runtime configurable rather than hard-coded?

mogoodrich:
I'd love to get rid of the Mirebalais application name, but unfortunately I forget how much of a lift it is, there might be a ticket (and worth ticketing if not).

I think the main thing would be to set up some sort of alias... I think the main risk is breaking everyones bookmarks, etc....

mseaton:
It's a good point.  I'd rather have something hacky for mirebalais that adds something to the Apache config or mod_jk config or whatever that resolves /mirebalais and /openmrs both to /openmrs.  Maybe that is easily doable.  @brandones ?

brandones:
> any chance that MFE can support this being more runtime configurable rather than hard-coded?

Yeah theoretically, though I'd rather we not. This is how the SPA Module used to be structured. It carried its own index.html, which it could interpolate with global properties. In the new system, the index.html file is provided in the application data directory. So really this new system is in a way more dynamic/flexible—the index.html file can be changed at will—it's just that the server doesn't have any control over it.

mseaton:
OK

brandones:
>  I'd rather have something hacky for mirebalais that adds something to the Apache config or mod_jk config or whatever that resolves /mirebalais and /openmrs both to /openmrs. Maybe that is easily doable. @brandones ?

Just to repeat what I've said on calls, for posterity: Yes, absolutely, let's do that sometime.

PR #139, opened 2021-05-28
Add pih repo url basing on OS version

jmbabazi:


mirebalais-modules/openmrs/manifests/init.pp
```diff
@@ -36,6 +36,9 @@
     $lacolline_username = decrypt(hiera('lacolline_username')),
     $lacolline_password = decrypt(hiera('lacolline_password')),
 
+    # os version
+    $old_os_version = hiera('old_os_version'),
```
mogoodrich, line 40: if this is only for ubuntu 14, might be clearer to name the variable "ubuntu_14" or something like that?

mogoodrich (approved):
@jmbabazi LGTM, just one suggestion, thanks!

PR #143, opened 2021-07-16
UHM-5643 Release of ZL 1.25.0-SNAPSHOT and 1.24.0

lnball:
Can you confirm this is correct for ZL?


jmbabazi:
@lnball , this looks god to me merging it in!

PR #147, opened 2021-08-23
PETL-94: Set different run times for petl jobs based on a specific se…

mogoodrich:
…rver

### DIFF
--- a/hieradata/common.yaml
+++ b/hieradata/common.yaml
@@ -87,13 +87,14 @@ pwa_filename: ''
 
 # PETL configuration
 petl: "petl"
-petl_version: "2.2.0"
+petl_version: "2.5.0-SNAPSHOT"
 petl_java_home: "/usr/lib/jvm/java-8-openjdk-amd64"
 petl_java_opts: ""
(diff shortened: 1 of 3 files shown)


hieradata/common.yaml
```diff
@@ -106,7 +107,7 @@ petl_sqlserver_databaseName: openmrs_extractions
 petl_sqlserver_user: foo
 petl_sqlserver_password: bar
 petl_error_subject: Petl errors
-petl_cron_hour: '*/6'
```
mogoodrich, line 109: I renamed this just to clearly distinguish it from the new cron I added

mirebalais-modules/petl/templates/application.yml.erb
```diff
@@ -3,6 +3,8 @@ petl:
   homeDir: "/home/<%= @petl %>"
   datasourceDir: "<%= @petl_datasource_dir %>"
   jobDir: "<%= @petl_job_dir %>"
+  schedule:
+    cron: "<%= @petl_schedule_cron %>"
```
mogoodrich, line 7: This will apply this schedule to any jobs that don't have a custom schedule defined in their job.yml

PR #150, opened 2021-09-28
Haiti HIV configs

jmbabazi:


manifests/site.pp
```diff
@@ -62,6 +62,33 @@
...
+
+  include java
+  include mysql_setup
+  include apache2
+  include tomcat
+
+  include openmrs
+  include openmrs::initial_setup
+
+  include percona
```
mogoodrich, line 86: Are we still using Percona on these machines?  (Maybe we are, and I'm just forgetting that... :)

mogoodrich (approved):
One question, but generally LGTM!

manifests/site.pp
jmbabazi: We aren't! I will disable that.

Thanks

PR #151, opened 2021-09-28
UHM-5787

mogoodrich:


hieradata/zlemr.pih-emr.org.yaml
```diff
@@ -2,7 +2,7 @@
 webapp_name: openmrs
 site_domain: zlemr.pih-emr.org
 
-pih_config: haiti-hiv,haiti-hiv-ci
+pih_config: haiti,haiti-hiv
```
mogoodrich, line 5: I removed the "haiti-hiv-ci" here since this will be a production server so we don't want any of the CI-specific functionality.

jmbabazi: LGTM

PR #155, opened 2021-10-14
Adding password and backup folder path to ces yaml file

jmbabazi:


jmbabazi:
Hi @brandones , you meant for the servers located in the mexico, why just capitan? Or you meant for all mexico servers why just capitan?

Anyway, I just followed what you or Joseph had initially setup. capitain is on the nodes for the servers in the cloud, the two servers had the right configuration but capitan didn't, It was picking up the vars in common.yaml see https://github.com/PIH/mirebalais-puppet/blob/1d3a629c64c190df480286ed014359ac028e18c8/manifests/site.pp#L410

If you want all the servers to have this, then we should enable "openmrs::backup" on the other nodes and ensure that we have the rights vars in each respective yaml file.

PR #159, opened 2021-10-25
Adding petl installation for Malawi DW

jmbabazi:


jmbabazi:
This has been handled on another branch

PR #161, opened 2021-11-03
Disable PETL UHM-5905

jmbabazi:


mseaton (approved):
LGTM @jmbabazi

hieradata/petl-test.pih-emr.org.yaml
```diff
@@ -10,6 +14,11 @@ petl_sqlserver_databaseName: openmrs_petl_test
 petl_database_url: "jdbc:h2:mem:petl"
 petl_error_subject: Petl-Test petl errors
 petl_cron_hour: '*/4'
+petl_config_dir_url: "https://github.com/PIH/pih-pentaho"
```
mseaton, line 17: This isn't really petl_config_dir_url, this is the pentaho job code that is used by Malawi, Rwanda, and Liberia.  I'd make it clear that this is not something that is part of a standard "petl" install, but is a nuance for how we need to pull in all of the job configuration files for those countries.

hieradata/petl-test.pih-emr.org.yaml
```diff
@@ -10,6 +14,11 @@ petl_sqlserver_databaseName: openmrs_petl_test
...
+petl_config_name: pih-pentaho
```
mseaton, line 18: Same here with petl_config_name.  Right now, this is just pentaho code.  The petl job files aren't anywhere yet.  We still need to figure out where all of these files will be hosted, packaged, deployed, and pulled in.  That's this ticket:  MLW-1261

hieradata/common.yaml
```diff
@@ -89,6 +89,8 @@ pwa_filename: ''
 
 # PETL configuration
 petl: "petl"
+petl_user: "petl"
+petl_base_dir: "petl"
 petl_version: "2.6.0"
```
mseaton, line 95: This needs to be 3.0.0-SNAPSHOT

hieradata/petl-test.pih-emr.org.yaml
```diff
@@ -1,6 +1,10 @@
 ---
 site_domain: petl-test.pih-emr.org
 
+# config files
+config_name: openmrs-config-zl
+config_version: 1.27.0
```
mseaton, line 6: We probably want to build this around the idea that all countries will build and publish a particular artifact that contains their "etl config", and which is a zip file that contains "jobs", "datasources", application.yml template potentially (and which basically can be unzipped into the petl home directory.  This doesn't exist yet, but we might want to expedite this - this is related to MLW-1261 and having some sort of standardization.

hieradata/petl-test.pih-emr.org.yaml
```diff
@@ -10,6 +14,11 @@ petl_sqlserver_databaseName: openmrs_petl_test
 petl_database_url: "jdbc:h2:mem:petl"
 petl_error_subject: Petl-Test petl errors
 petl_cron_hour: '*/4'
+petl_config_dir_url: "https://github.com/PIH/pih-pentaho"
+petl_config_name: pih-pentaho
+petl_job_dir: "/home/petl/jobs"
```
mseaton: I'm not really sure I understand the purpose of this variable.  The jobs directory should always be called "jobs" and be a subdirectory of PETL_HOME for a given install.

mirebalais-modules/petl/manifests/init.pp
```diff
@@ -170,4 +173,4 @@
     require => File["/usr/local/sbin/checkPetlErrors.sh"]
   }
 
-}
+}
```
mseaton, line 176: All of this generally seems right, that we would by default install a single petl instance on a server at /opt/petl, but could allow for multiple instances (eg. for petl-test server), by installing into /opt/malawi-petl, opt/haiti-petl, etc.  And then within those directories, the standard setup of jobs, datasources, bin/petl.jar, bin/application.yml, petl.conf, etc.

mseaton (approved):
Generally looks good @jmbabazi .  I think we will need to evolve how we pull in our job configurations over time (eg. right now we need to pull in the source from pih-pentaho, and also jobs from config projects and elsewhere, but ideally we can evolve to a model where this is more standardized).  That can happen later though, assuming adjusting this is straightforward when the time comes.

mseaton (approved):
Looks fine @jmbabazi .  I was hoping we could have different application.yml.erb files rather than a big if/then statement, but we can evolve this over time as needed I guess.  Ultimately, this just needs to be in the format that works best for you to maintain.

mogoodrich (approved):
LGTM, thanks @jmbabazi ... I do agree with @mseaton  about evolving to separate application.yml files as this might be easier to maintain.  Not sure if this would work work, but I'm wondering if the "petl_site" variable could be used here to load different files based on site:

https://github.com/PIH/mirebalais-puppet/blob/master/mirebalais-modules/petl/manifests/init.pp#L111

(ie something like "content => template("petl/application-$petl_site.yml.erb")"

jmbabazi:
> application.yml.erb

I've changed this to have site specific application.yml files

mogoodrich:
Oh, cool, so I guess that suggestion I made worked?

PR #165, opened 2021-12-07
Enabling Malawi openmrs installation using puppet

jmbabazi:


mirebalais-modules/openmrs/manifests/apzu.pp
```diff
@@ -0,0 +1,75 @@
...
+    group   => $tomcat,
+    mode    => '0644',
+    require => File["${tomcat_home_dir}/.OpenMRS"]
+  }
+
+  exec { 'cleanup-malawi-modules-dir':
+    command => "rm -rf /tmp/malawi-modules && rm -rf /tmp/malawi-distro && rm -rf /tmp/malawi-distribution"
+  }
+
+  wget::fetch { 'download-malawi-modules':
```
mseaton, line 43: This doesn't seem right.  Didn't we move to pull in the new distribution artifact (that has war, modules, etc), rather than just a modules zip?

mirebalais-modules/openmrs/manifests/apzu.pp
```diff
@@ -0,0 +1,75 @@
...
+    source      => "${repo_url}:81/malawi-repo/modules.zip",
+    destination => "/tmp/malawi-modules",
+    timeout     => 0,
+    verbose     => false,
+    redownload  => true,
+    require     => Exec['cleanup-malawi-modules-dir']
+  }
+
+  exec { 'extract-malawi-modules':
+    command => "unzip -o /tmp/malawi-modules -d /tmp/malawi-distro",
```
mseaton, line 53: Same as above.

mirebalais-modules/openmrs/manifests/apzu.pp
```diff
@@ -0,0 +1,75 @@
...
+    require => [ Wget::Fetch['download-malawi-modules'], Package['unzip']]
+  }
+
+  exec { 'move-malawi-modules':
+    command => "mv /tmp/malawi-distro/* /tmp/malawi-distribution && rm -rf ${tomcat_home_dir}/.OpenMRS/modules/* && cp -r /tmp/malawi-distribution/openmrs_modules/* ${tomcat_home_dir}/.OpenMRS/modules/ && chown -R ${tomcat}:${tomcat} ${tomcat_home_dir}/.OpenMRS/modules",
```
mseaton, line 58: This seems better, in that you are doing modules and war, but the naming throws me off, and I still don't see where you are actually downloading the distribution.

mseaton (commented):
See a few comments @jmbabazi .  Overall looking good.

hieradata/petl-test.pih-emr.org.yaml
```diff
@@ -6,15 +6,15 @@ server_timezone: America/New_York
 config_name: openmrs-config-zl
 config_version: 1.28.0-SNAPSHOT
 ### note this version goes hand in hand with the config_version
-petl_config_version: 1.28.0-SNAPSHOT
+petl_config_version: "%{hiera('config_version')}"
```
mogoodrich, line 9: Is this to make sure that petl_config_version always equals config_version?  If they should always be the same, should we consider removing the petl_config_version altogether, or are there places where they need to be different?

hieradata/humci.pih-emr.org.yaml
```diff
@@ -6,6 +6,8 @@ pih_config: mirebalais,mirebalais-humci
 config_name: openmrs-config-zl
 config_version: 1.28.0-SNAPSHOT
 
+tomcat_home_dir: /home/tomcat7
```
mogoodrich, line 9: Are there any sites currently where Tomcat isn't in /home/tomcat7?  Should this default be set in the common.yaml file?

hieradata/common.yaml
```diff
@@ -12,6 +12,8 @@ site_domain: foo.org
 package_release: 'unstable/'
 package_version: '1.0.6088-1'
 
+repo_url: ENC[guOgejmsDLkHjo2iCrV9MAWEt/Y9usi32Rs4qPepBu0=]
```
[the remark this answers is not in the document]
mogoodrich: So it looks this bamboo.pih-emr.org and we are encrypting that?  Not sure that is necessary, but I guess it can't hurt.

mirebalais-modules/openmrs/manifests/init.pp
```diff
@@ -1,51 +1,9 @@
 class openmrs (
-    $openmrs_db = hiera('openmrs_db'),
```
mogoodrich, line 2: this all got moved into manifests/pihemr.pp, correct?

That makes sense, though why is anything left in this file?

mogoodrich (commented):
@jmbabazi added some comments, thanks!

hieradata/common.yaml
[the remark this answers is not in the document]
jmbabazi: on this same place, we host our ci which requires login. I think we should work on minimizing the exposition to this since its more used internally. Encrypting the url is not the best way but its a start to a conversation.  We can discuss on having these modules and other artifacts hosted somewhere and not on this specific system.

manifests/malawi.pp
```diff
@@ -1,22 +1,23 @@
 Exec { path => [ '/bin/', '/sbin/' , '/usr/bin/', '/usr/sbin/', '/usr/local/bin/' ] }
 
-node default {
+node 'neno.pih-emr.org' {
```
cioan, line 3: @jmbabazi , can you please use a node name that actually works? neno.pih-emr.org is the production server and we are not going to run this puppet script on that server. you could use the test server name?

mirebalais-modules/openmrs/manifests/apzu.pp
jmbabazi: thanks for the catch, this will be fixed.

hieradata/humci.pih-emr.org.yaml
jmbabazi: I think Malawi upper neno runs two tomcat instances. If we the tomcatDir to common.yaml, then we give puppet that flexibility of installing the two tomcat instances (though we still need to work on how the tomcat ports will be handled in server.xml and other apache configurations)

hieradata/petl-test.pih-emr.org.yaml
jmbabazi: @mogoodrich, we have https://github.com/PIH/apzu-etl which is the petl repo for malawi. This repo is only for the malawi etl not for other configurations. If we use config_version wouldn't it be confusing for other people?

manifests/malawi.pp
jmbabazi: Changed that to default @cioan

jmbabazi (commented):
@mseaton @mogoodrich @cioan please review my new changes. Made the changes you requested for and made class openmrs () the common class

mirebalais-modules/openmrs/manifests/apzu.pp
```diff
@@ -38,7 +38,7 @@
   }
 
   wget::fetch { 'download-malawi-openmrs-distribution':
-    source      => "${repo_url}:81/malawi-repo/${pih_malawi_distribution_version}.zip",
+    source      => "${repo_url}:81/malawi-repo/pihmalawi-distribution-${pih_malawi_distribution_version}.zip",
```
cioan, line 41: @jmbabazi , here is how I got wget to download the distribution zip regardless of version: 
`wget -r -l1 --no-parent -A "pihmalawi-distribution*.zip" http://bamboo.pih-emr.org/malawi-repo/ -O pihmalawi-distribution.zip`

jmbabazi: Thanks @cioan, not sure if that will work since we are using puppet module. Do you us to change this to a command?

mseaton (approved):

hieradata/humci.pih-emr.org.yaml
mogoodrich: Cool... thanks @jmbabazi to be clear I think it's fine to keep this "tomcat_home_dir" variable if there will be servers that have a non-standard install directory, but just that it probably makes sense to put the default in "common.yaml:.

hieradata/petl-test.pih-emr.org.yaml
mogoodrich: I think I understand... yes, if they are two separate things (and might be different in some cases), could to keep them separate like this.

mirebalais-modules/openmrs/manifests/init.pp
mogoodrich: It's still unclear to me if this file is ever used at all now?

mogoodrich (approved):
@jmbabazi  I was still a little confused by one small thing, but generally works for me... I defer to @cioan  and @mseaton  about the specifics of the Malawi install.

mirebalais-modules/openmrs/manifests/apzu.pp
cioan: No need to change this. I thought that you might have run into this problem with downloading a file with wild characters. It took me a little searching on Google to get this syntax correct and I offered it here.
We just have to remember to update this version when we release a new version of pihmalawi. Do you use a distro version variable for the other puppet installs? I mean, every time a new pihemr version is release, do you have to go into the puppet script and update some variable to the new version?

jmbabazi: Thanks @cioan , for the pihemr we use the debian packages, so we usually update the package version once we cut a gold https://github.com/PIH/mirebalais-puppet/blob/96fec417261b77780615d86dff49db395f7467bf/hieradata/common.yaml#L13. We also have another variable for the config_version

cioan: Thanks @jmbabazi ! This looks good to me!

mirebalais-modules/openmrs/manifests/init.pp
jmbabazi: It is still used @mogoodrich , in other files we have "require openmrs" line which calls for tasks in this file to be executed. This file can hold all common tasks, This file is required for pihemr.pp, apzu.pp and backup.pp

PR #169, opened 2022-01-07
UHM-6054: O3: Config not being loaded on humci

brandones:


mirebalais-modules/openmrs/manifests/pihemr.pp
```diff
@@ -180,12 +180,18 @@
 
     exec{'install-openmrs-frontend':
       command => "rm -rf /tmp/frontend && unzip -o /tmp/${frontend_name}.zip -d /tmp/frontend && rm -rf /home/${tomcat}/.OpenMRS/frontend && mkdir /home/${tomcat}/.OpenMRS/frontend && cp -r /tmp/frontend/*/* /home/${tomcat}/.OpenMRS/frontend",
-      require => [ Wget::Fetch['download-openmrs-frontend'], Package['unzip'], File["/home/${tomcat}/.OpenMRS"] ],
-      notify  => [ Exec['tomcat-restart'] ]
```
mseaton, line 184: Will removing this notify to restart tomcat fix the CI issues with server restarts that we have been experiencing?  @mogoodrich FYI?

mseaton (approved):
Looks good to me @brandones - thanks.  I don't know if this is the optimal way we want things setup down the road, but if it fixes the current issues all looks good.

mirebalais-modules/openmrs/manifests/pihemr.pp
mogoodrich: It's certainly a possibility... the issue is occurring during a tomcat restart... was this notify added just before Christmas @brandones ?  (That being said, I don't understand *why* this would be causing a problem, but I've been pretty flummoxed about this one...)

mogoodrich (approved):
LGTM, thanks @brandones !

mirebalais-modules/openmrs/manifests/pihemr.pp
brandones: Yeah I wouldn't expect this to actually change anything—I think Tomcat will still always restart. But now, in some formal sense, Puppet knows it doesn't have to restart Tomcat if nothing but the frontend was updated.

PR #172, opened 2022-01-13
PETL-114: Update HUM-CI to use SQL Server as datasource for petl_job_…

mseaton:
…execution


mogoodrich (commented):
this generally looks good to me, but I was surprised not to see a reference to the "petl_database_url" that we define in the various yml files to set whether to use a file-based or memory-based H2 database.

This suggested that variable is no longer used, or am I missing something?  I'm fine if that's the case but let's just remove that variable from the yml files to avoid confusion.

mseaton:
@mogoodrich - https://github.com/PIH/mirebalais-puppet/pull/173

PR #174, opened 2022-01-13
PETL-114: Set Malawi, CES, CES-TEST, ZL-REPORT, and ZL-TEST to use SQ…

mseaton:
…L Server for Job Execution Status.  Also change default logging to INFO.


mogoodrich (commented):
LGTM!

PR #175, opened 2022-01-14
UHM-6073 - PETL deploy should stop PETL before updating configuration

mseaton:


jmbabazi:
LGTM @mseaton , merging it in!

mogoodrich:
Makes sense to me, @mseaton ... while reviewing I noticed what I think was a missing dependency, and added it (and a comment):
https://github.com/PIH/mirebalais-puppet/commit/533817df1e723ff25af1b75d0b060ab69cb10fed

PR #178, opened 2022-01-18
Mlw 1298

jmbabazi:


hieradata/malawi-dw.pih-emr.org.yaml
```diff
@@ -18,6 +18,11 @@ petl_mysql_user: ENC[+DF7mUCIuCRoWd2GgG8mvQ==:FFIYj69sx/ENFvq6fnQ18A==:K1hGYTFLM
 petl_mysql_password: ENC[spcYc6R/l+JkjPm5P//qw0LVQCdrD8zEfIEZr/AJ4rU=]
 petl_mysql_options: "autoReconnect=true&sessionVariables=default_storage_engine%3DInnoDB&useUnicode=true&characterEncoding=UTF-8&serverTimezone=US/Eastern"
 
+# PETL database for Apzu
+install_petl_warehouse_db: true
+petl_warehouse_db: openmrs_warehouse
```
cioan, line 23: @jmbabazi , we are not installing the Pentaho openmrs_warehouse database on the Malawi DW server. Only on Lisungwi and Upper Neno EMR servers.

hieradata/malawi-dw.pih-emr.org.yaml
```diff
@@ -18,6 +18,11 @@ petl_mysql_user: ENC[+DF7mUCIuCRoWd2GgG8mvQ==:FFIYj69sx/ENFvq6fnQ18A==:K1hGYTFLM
...
+petl_mysql_user_ip: localhost
+
 # PETL Sql Service
 petl_sqlserver_host: 10.160.22.17
```
cioan, line 27: we should not be using IP addresses here.

cioan (changes requested):

hieradata/malawi-dw.pih-emr.org.yaml
jmbabazi: @cioan, why shouldn't we use IP? 

The IP is for the sql instance and its a vpn IP.

hieradata/lisungwi.pih-emr.org.yaml
```diff
@@ -23,12 +23,20 @@ petl_config_version: 1.0.0-SNAPSHOT
...
 
+petl_cron_time: "0 0 2 ? * *"
+petl_server_port: 8181
+
+# PETL database for Apzu
+install_petl_warehouse_db: true
+petl_warehouse_db: openmrs_warehouse
+petl_mysql_user_ip: localhost
+
+install_petl_warehouse_db: true
```
cioan, line 38: It looks like lines 38-40 are duplicates?

cioan (changes requested):

cioan (approved):
Thanks @jmbabazi !

PR #184, opened 2022-02-09
UHM-6162 Release openmrs-frontend-pihemr

brandones:


mogoodrich (approved):
It's likely preferably to have the release version (1.0.0) in the common.yaml, and have the staging servers override with the 1.1.0-SNAPSHOT, instead of vice versa, to avoid rolling out a snapshot to a prod server by accident, but I realize there more staging servers than prod servers, so I'm fine with doing it this way.

brandones:
Yeah before long I would like to have separate frontend packages for different sites anyway. https://pihemr.atlassian.net/browse/UHM-6171

PR #187, opened 2022-02-14
Enabling J9 on hinche-ci and ci

jmbabazi:


hieradata/ci.pih-emr.org.yaml
```diff
@@ -1,7 +1,7 @@
 ---
 site_domain: ci.pih-emr.org
 
-pih_config: haiti,haiti-hsn,haiti-hsn-ci
+pih_config: haiti,haiti-hinche,haiti-hinche-ci
```
mogoodrich, line 4: I like your idea of using Hinche-CI for this, I had forgotten about it... so maybe we don't have to change ci.pih-emr.org.yaml at all?

mogoodrich (commented):
I think we can just use hinche-ci.pih-emr.org and you can ignore my comments about changing ci.pih-emr.org...

hieradata/ci.pih-emr.org.yaml
jmbabazi: thanks @mogoodrich , change back to haiti-hsn

PR #188, opened 2022-02-17
UHM-6171: Move ZL servers to frontend-zl package

brandones:
See https://s01.oss.sonatype.org/content/repositories/snapshots/org/pih/openmrs/openmrs-frontend-zl/
and https://s01.oss.sonatype.org/content/repositories/releases/org/pih/openmrs/openmrs-frontend-zl/

https://pihemr.atlassian.net/browse/UHM-6171

### DIFF
--- a/hieradata/ami.pih-emr.org.yaml
+++ b/hieradata/ami.pih-emr.org.yaml
@@ -7,7 +7,7 @@ config_name: openmrs-config-zl
 config_version: 1.34.0-SNAPSHOT
 
 frontend_name: "openmrs-frontend-zl"
-frontend_version: 1.2.0-SNAPSHOT
+frontend_version: 1.1.0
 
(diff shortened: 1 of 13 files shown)


mseaton (approved):
Looks good to me.  Seems a bit silly that we don't have some common "haiti.yaml" file that all of the ZL servers can inherit from and override if needed.  @jmbabazi is this normal?

brandones:
Oh something similar for Mexico would be awesome, also. Thanks @mseaton !

mogoodrich:
@brandones  @jmbabazi  @mseaton  I'm forgetting the details, but years ago I believe support was removed (?) from Puppet for nested levels of inheritance in yam files, so you could only have a "common.yaml" and then a site-specific one (and therefore can't do what you are suggesting with a haiti.yml file @mseaton ) .  I forget all the reasoning behind this, but this was literally back in 2014 or 2015 so there's probably a better way to support this kind of thing in more recent versions of Puppet.

PR #190, opened 2022-02-24
Uhm 6226 openmrs mail

jmbabazi:


mirebalais-modules/openmrs/manifests/pihemr.pp
```diff
@@ -114,7 +119,16 @@
     content => template('openmrs/openmrs-runtime.properties.erb'),
     owner   => $tomcat,
     group   => $tomcat,
-    mode    => '0644',
+    mode    => '0600',
+    require => File["${tomcat_home_dir}/.OpenMRS"]
+  }
+
+  file { "${tomcat_home_dir}/.OpenMRS/mail.properties":
```
mseaton, line 126: The mail.properties file is really just meant to be for testing.  We can use it, but the expected way is just to put these properties directly into the runtime properties file.

mirebalais-modules/openmrs/templates/mail.properties.erb
```diff
@@ -0,0 +1,12 @@
+mail.debug=false
+mail.default_content_type=text/plain
```
mseaton, line 2: Do we want text/plain, or text/html?

mirebalais-modules/openmrs/templates/mail.properties.erb
```diff
@@ -0,0 +1,12 @@
...
+mail.from=<%= @smtp_username %>
+mail.password=<%= @smtp_userpassword %>
+mail.smtp.auth=true
```
mseaton, line 5: I thought you said that we'd typically be using port 25 and sendmail, not smtp and gmail?  Let's make these match what we expect to use.

hieradata/common.yaml
```diff
@@ -171,7 +172,7 @@ apache_cron_restart_min: 03
 az_secret: ENC[1oQfg4Gi51CDU3XTOIkHatJn2W+vAxx1Juj/wYnVwyRtEFk7hm7XAkA0CbKlsUcUdED4rBiSDIYAlTiO3aBryg==]
 az_url: ENC[6NweILo1Eq7pKNo88HD1o/fyAWOijtE/7o3Pn+pd8tJ9q2Z1PerwvhtXMsVi2nl37EjQL96k8gApwQCE+nKnaw==]
 
-java_memory_parameters: '-Xmx2048m -Xms1024m'
+java_memory_parameters: '-Xmx4096m -Xms1024m'
```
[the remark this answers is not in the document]
jmbabazi: Yes we do @mseaton!

mirebalais-modules/openmrs/manifests/initial_setup.pp
```diff
@@ -57,10 +57,17 @@
     refreshonly => true
   }*/
 
+  # hack to let us remote the mirebalais metadata module from the build; can be removed after it has been removed from all servers
+  exec { 'make mirebalais metadata module not mandatory - error running this command can be ignored when provisioning new server':
+    command     => "mysql -u${openmrs_db_user} -p'${openmrs_db_password}' ${openmrs_db} -e 'update ${openmrs_db}.global_property set property_value=\"false\" where property=\"mirebalaismetadata.mandatory\"'",
+    user        => 'root',
+    require => Package[$package_name],
+  }
+
```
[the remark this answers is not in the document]
jmbabazi: Unrelated to mail! I removed the notify tomcat since it results into failed dependency when we install new instances. Eventually we will remove these lines but for now, not notifying tomcats allows for new installations

mirebalais-modules/openmrs/manifests/pihemr.pp
jmbabazi: I think we should this approach. Especially if we plan to expand mail to other instances not only the PIHEMR

mirebalais-modules/openmrs/templates/mail.properties.erb
jmbabazi: localhost and port 25 works too but I think this is better since its the SSL port. Lets use this

mirebalais-modules/openmrs/templates/mail.properties.erb
jmbabazi: html is better. Changed that

PR #193, opened 2022-03-01
Uhm 6091 Reconfigure set-up Data Pipeline for CES on PETL-Test

jmbabazi:
https://pihemr.atlassian.net/browse/UHM-6091


mirebalais-modules/petl/manifests/init.pp
```diff
@@ -15,200 +15,199 @@
...
-
-    cron { 'petl-error':
-      ensure  => present,
-      command => "/usr/local/sbin/petl-${petl_site}-checkErrors.sh &> /dev/null",
-      user    => 'root',
-      hour    => "${petl_check_errors_cron_hour}",
-      minute  => 00,
-      require => File["/usr/local/sbin/petl-${petl_site}-checkErrors.sh"]
-
+  $array_list = [0, 1]
```
mogoodrich, line 31: Might be worth adding a comment here that we need to increase this number if we end up supporting arrays with more than two elements?  

I don't know the Puppet/Ruby syntax, but I also suspect might just be able to do "each($petl) |$index| {"  instead of creating the $array_list, which I would assume mean "iterate over each element in the $petl array, passing in the index of the element into the variable $index...

mirebalais-modules/petl/manifests/java.pp
```diff
@@ -1,6 +1,11 @@
...
 ) {
   package { "openjdk-8-jdk":
-    ensure  => present
+    ensure => present
   }
-}
+
+  exec { "Hack to support multiple petl service on one server":
+    command => "cp -r /usr/lib/jvm/java-8-openjdk-amd64 /usr/lib/jvm/java-8-openjdk-amd64-2",
+    require => Package['openjdk-8-jdk'],
```
mogoodrich, line 9: We can't run two PETL instances using the same JVM?

mogoodrich (commented):
A few comments... it seems strange to me that you need to clone the jvm and have a separate one for each Petl instance? (If I'm following the code correctly) But generally looks good, thanks @jmbabazi !

mirebalais-modules/petl/manifests/init.pp
mseaton: +1

mirebalais-modules/petl/manifests/java.pp
mseaton: Right, I'd expect we would just have the 1 java installation, and both would use it.

mirebalais-modules/petl/manifests/init.pp
```diff
@@ -15,200 +15,219 @@
...
+        $petl_repo = "snapshots"
+      }
+      else {
+        $petl_repo = "releases"
+      }
+
+      wget::fetch { "download-petl-jar for ${petl_site[$index]}":
+        source      => "https://s01.oss.sonatype.org/service/local/artifact/maven/content?g=org.pih&a=petl&r=${petl_repo
+          }&p=jar&v=${petl_version}",
+        destination => "${petl_home_dir[$index]}/bin/petl-${petl_version}.jar",
```
mogoodrich, line 92: Try changing this to:
destination => "${petl_home_dir[$index]}/bin/${petl[$index]}-${petl_version}.jar"

So that the jar ends up having the same name as the service...

mirebalais-modules/petl/manifests/init.pp
```diff
@@ -15,200 +15,219 @@
...
+        timeout     => 0,
+        verbose     => false,
+        redownload  => ('SNAPSHOT' in $petl_version),
+        require     => File["${petl_home_dir[$index]}/bin"]
+      }
+
+      file { "${petl_home_dir[$index]}/bin/petl-${petl_version}.jar":
```
mogoodrich, line 99: And then you'd probably need to update this the same way as above... (and maybe some other things further along as well, I haven't checked further...)

jmbabazi: So I tried this but I got the same insserv error. I decided to use symbolic links and it worked. I think using symbolic links is hacky but it gets the work done, And I think we may need to rethink supporting multiple petl instances on the same server.

mirebalais-modules/petl/templates/application-ces-test.yml.erb
```diff
@@ -1,7 +1,7 @@
 petl:
-  homeDir: "<%= @petl_home_dir -%>"
-  datasourceDir: "<%= @petl_home_dir -%>/<%= @petl_config_dir -%>/datasources"
-  jobDir: "<%= @petl_home_dir -%>/<%= @petl_config_dir -%>/jobs"
+  homeDir: "/opt/ces-test-petl"
```
mogoodrich, line 2: I would think that on servers with only one version of Petl we'd want to stick to the default install directory of /opt/petl?  Or will that not work and/or am I not thinking of a reason why it would be better to do it this way? (and same question for the other examples below where we do the same)

mogoodrich (commented):
One question, @jmbabazi , but generally looks good, thanks!

mirebalais-modules/petl/templates/application-ces-test.yml.erb
jmbabazi: thanks @mogoodrich, I agree. will change these to /opt/petl

PR #200, opened 2022-03-31
UHM-6376 hinche-ci and zltraining are both configured to consume zlem…

mseaton:
…r ids from zanmi


hieradata/zltraining.pih-emr.org.yaml
```diff
@@ -28,8 +28,6 @@ mysql_server_id: 1
 mysql_innodb_buffer_pool_size: 3G
 mysql_innodb_buffer_pool_instances: 3
 
-openmrs_auto_update_database: true
```
mseaton, line 31: Duplicate property in file

PR #212, opened 2022-06-05
Update apzu-etl snapshot-version

kmatiya:
Version has been updated as the apzu-etl(https://github.com/PIH/apzu-etl) pom version has been updated to allow puppet to load the updated data warehouse tables on the mssql reporting server.

cc:
@jmbabazi


jmbabazi:
No necessary!

PR #214, opened 2022-07-27
Update to latest gold

brandones:


hieradata/common.yaml
```diff
@@ -11,7 +11,7 @@ site_domain: foo.org
 
 package_name: pihemr
 package_release: 'unstable/'
-package_version: '1.0.6446-1'
+package_version: '1.0.6484-1'
```
mseaton, line 14: Defer to @mogoodrich and others - do we do this in common.yaml?  I would think we'd only want to release a new gold where we are ready for it?

brandones:
@mseaton @mogoodrich  Changed this to only be for CES.

Joseph-Developer (approved):

mogoodrich (approved):
LGTM!

hieradata/common.yaml
mogoodrich: Generally, when we do a new Gold, we just update the version in the common.yml, but if we cut this gold just for Mexico, it's fine to just update the version number here for each Mexico server, and then update the common.yml when/if we want to release it everywhere.

PR #216, opened 2022-08-01
UHM-6576: Update to use JFrog URL for PIH EMR releases

mogoodrich:
This changes Puppet to point to JFrog as the apt-source for the PIH EMR.

It is untested, wanted others to take a look first, and then I can test it on ci.pih-emr.org.  We won't want to merge this in until we are ready to do the full switchover to the new Bamboo.


hieradata/emr.hum.ht.yaml
```diff
@@ -4,8 +4,6 @@ az_activitylog_backup_folder_path: activitylog/zl/mirebalais
 
 package_release: 'stable/'
```
mogoodrich, line 6: I don't *think* we need a custom URL here anymore for emr.hum.ht, but we will need to make sure we test/monitor next time we upgrade Mirebalais

mirebalais-modules/openmrs/manifests/pihemr.pp
```diff
@@ -51,30 +51,17 @@
...
-      repos       => '',
-      include_src => false,
-    }
+
+  apt::source { $package_name:
+    ensure      => present,
+    location    => "[trusted=yes] ${pihemr_repo_url}/${package_name}-repo",
+    release     => $package_release,
+    repos       => '',
+    include_src => false,
```
mogoodrich, line 64: I'm hoping we aren't going to need the old Ubuntu 14.04 branch, but we *are* running 14.04 in Mirebalais so we will need to do a little testing once we are ready to upgrade Mirebalais.

hieradata/common.yaml
```diff
@@ -17,6 +17,8 @@ config_name: ""
 config_version: ""
 
 repo_url: ENC[guOgejmsDLkHjo2iCrV9MAWEt/Y9usi32Rs4qPepBu0=]
+pihemr_repo_url: https://openmrs.jfrog.io/artifactory/deb-pih/
```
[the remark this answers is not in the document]
mseaton: At least for Bwenzi, I have updated it to pull from Maven for the APZU PWA, and I believe our release process typically is to run scripts on Bwenzi that transfer what is there to either Neno or Lisungwi, so I don't think we need to retain this for Malawi @mogoodrich .  @cioan ?

hieradata/emr.hum.ht.yaml
mseaton: Agreed, I don't see why we should.

mirebalais-modules/openmrs/manifests/pihemr.pp
```diff
@@ -24,7 +24,7 @@
   # PIH EMR config
   $config_name     = hiera('config_name'),
   $config_version  = hiera('config_version'),
-  $repo_url        = decrypt(hiera('repo_url')),
+  $pihemr_repo_url        = hiera('pihemr_repo_url')),
```
mseaton, line 27: Regardless of whether anything should still use "repo_url", I think the change to "pihemr_repo_url" is good.  Though maybe I'd change it to "pih_debian_repo_url".

mseaton (approved):

hieradata/common.yaml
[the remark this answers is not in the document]
mogoodrich: There are some places in the Malawi job that it's referenced, but it's entirely possible that it isn't used anymore... I can start a different PR for pulling that stuff out.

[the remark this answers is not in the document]
mogoodrich: (Or maybe I'll include it in this one, TBD...)

mirebalais-modules/openmrs/manifests/pihemr.pp
mogoodrich: I'm also curious if this will *replace* the old pihemr repo location, we will need to test.

hieradata/belladere.pih-emr.org.yaml
```diff
@@ -2,7 +2,7 @@
 webapp_name: openmrs
 site_domain: belladere.pih-emr.org
 
-package_release: 'stable/'
```
mogoodrich, line 5: Not sure if removing this / is necessary, but it isn't needed, so I removed it everywhere.

hieradata/common.yaml
```diff
@@ -10,13 +10,15 @@ pih_config: default
 site_domain: foo.org
 
 package_name: pihemr
-package_release: 'unstable/'
-package_version: '1.0.6446-1'
+package_release: 'unstable'
+package_version: '1.0.20220802174134-1'
```
mogoodrich, line 14: Beforee we merge this in, we will want to change this to whatever we are considering out latest stable gold on the new server

mirebalais-modules/openmrs/manifests/pihemr.pp
mogoodrich: Confirmed this does to a replace!

mogoodrich:
Note: remember when we merge this back in to set ci.pih-emr.org back to master

mseaton (approved):
This all looks good @mogoodrich .  Probably best to go ahead and make the change to package_version in common.yml back to the latest stable gold version to be safe now.  We could even take the deb that is our current gold and upload it to jfrog and point to that.

mogoodrich:
Good point @mseaton .  Once we merge in this change, we will need to package_version to point to a version in JFrog... I was assuming we'd just need to cut a new gold, but posting the old gold up to JFrog is probably the safest choice

PR #219, opened 2022-10-17
Adding commcare user

jmbabazi:


hieradata/kgh-test.pih-emr.org.yaml
```diff
@@ -7,3 +7,7 @@ config_version: 1.9.0-SNAPSHOT
 
 mysql_root_password: ENC[oWqIfmn/ER4CYIvjJSpIzQ==]
 openmrs_db_password: ENC[BNK3oX/1a+ta7Lal9jVd1A==]
+
+#commcaregit
```
mogoodrich, line 11: small typo here

mogoodrich (commented):
Thanks @jmbabazi ... addition of commcare user looks good, just flagged one small typo.

I see you are also removing a bunch of apache files here.. was this intentional?  It probably should happen in a separate commit for better tracking, since I don't think it's related to the commcare stuff?

mogoodrich:
Oh, @jmbabazi , my apologies, I see your comment on the ticket now regarding removing the certs... I should have reviewed that before starting to the review the PR.

I think think it might make sense to add those as a separate commit for clarity...

PR #220, opened 2022-10-18
Ensuring apache2 uses rewrite rule for accessing webapps

jmbabazi:


mirebalais-modules/apache2/manifests/init.pp
```diff
@@ -83,13 +83,15 @@
   }
 
   file { '/var/www/html/.htaccess':
-    ensure => file,
-    source => 'puppet:///modules/apache2/www/htaccess'
```
mogoodrich, line 87: Since we are removing this file, we want to remove it from the puppet project as well, unless we are using it elsewhere?

mirebalais-modules/apache2/manifests/init.pp
```diff
@@ -83,13 +83,15 @@
...
+  ensure => absent,
+  purge => true,
+  force => true,
   }
 
   file { '/var/www/html/index.html':
-    ensure => file,
-    content => template('apache2/index.html.erb')
```
mogoodrich, line 92: Same here... if we aren't using it elsewhere, let's remove the index.html.erb file from the Puppet project entirely.

mogoodrich (approved):
Thanks @jmbabazi ... my knowledge of this stuff is pretty limited, but this seems to be reducing the lines of configuration and simplifying things, so seems good to me.

I did add a couple lines about removing the the index.html and .htaccess files from Puppet if we are no longer using them.

Will also flag for Mike to review, since my knowledge is limited... :)  Thanks!

mseaton (approved):
LGTM at a glance, but really would need to test it out.

mirebalais-modules/apache2/manifests/init.pp
jmbabazi: Thank you. These are removed.

mirebalais-modules/apache2/manifests/init.pp
jmbabazi: Thank you. These are removed.

PR #221, opened 2022-10-19
re-enabling dumps for reporting tables

jmbabazi:


mirebalais-modules/mirebalais_reporting/templates/percona-openmrs-db-restore.sh.erb
```diff
@@ -25,6 +28,27 @@ echo cronjob started on $jobStartDate
...
+if [ $? -eq 0 ]; then
+        echo removing old reporting dump succesful
+    else
+        ## No hard exit here
+        echo failure removing old reporting dump
+        mail -s "$perconaSite: Failure removing old reporting dump" "$mailTo" < $perconaLogFile
+fi
+
+## creating new reporting dump file
+mysqldump -u${dbUser} -p${dbPass} openmrs reporting_report_design reporting_report_design_resource reporting_report_processor reporting_report_request scheduler_task_config scheduler_task_config_property serialized_object global_property > ${perconaHomeDir}/reporting/reporting_tables.sql
```
mseaton, line 44: We should discuss these tables and if they are the right ones.  I'm not sure why we need to preserve most of them.  My sense is that the only table we need to preserve is `reporting_report_request`.

`serialized_object`, `reporting_report_design`, `reporting_report_design_resource`, and `reporting_report_design_processor` are all related to the report _definitions_.  As far as I know, all of these definitions are being set up in the PIH EMR startup, and have consistent uuids across restarts.  We should not need to preserve these.  The exception would be if people are manually creating reporting definitions on the server and we want to preserve these, but I don't think we do?

`scheduler_task_config` and `scheduler_task_config_property` have nothing to do with reporting as far as I know.  I also don't think we need to preserve `global_property` for any reporting-related reason.  Are these here for some other purpose?  I'd recommend removing them and if there are reasons to add them back in, we can.

@dmdesimone / @mogoodrich do you know?

mirebalais-modules/mirebalais_reporting/templates/percona-openmrs-db-restore.sh.erb
```diff
@@ -87,6 +111,18 @@ else
         exit
 fi
 
+# restore reporting tables
+sleep 20
+mysql -u${dbUser} -p${dbPass} -D openmrs -e "SOURCE ${perconaHomeDir}/reporting/reporting_tables.sql;"
```
mseaton, line 116: I think we should first delete/truncate all rows from `reporting_report_request` that may have come in from the production server, and _then_ we should load in the backed up data from the `reporting_report_request` table from before the restoration.

What we are trying to achieve is to make sure that all of the historical report executions that have taken place on the reporting server (and which there may be data for in the application data directory) are still available for viewing/downloading in the UI after the restoration.  We don't want report executions that may have taken place on the production server (which are not related to reporting server needs, and which might have key clashes with the rows we are trying to restore).

mseaton (changes requested):

mirebalais-modules/mirebalais_reporting/templates/percona-openmrs-db-restore.sh.erb
mogoodrich: The only think I see currently in scheduler_task_config is the close stale visits tasks, so I don't see any strong need to preserve those tables.  This may have been from when we running some reports via the scheduler?

On the other hand, I don't remember the exact reasoning for copying the "global_property" table, but I do think it makes sense to preserve this one in case there is some global property based configuration we want to keep different between Reporting and Production (again, I think there was one, but I don't remember the details of it).

mogoodrich (commented):
Added a few comments, thanks @jmbabazi

mirebalais-modules/mirebalais_reporting/templates/percona-openmrs-db-restore.sh.erb
mogoodrich: Just to confirm ere there foreign key connections between the reporting_report_request and the other tables that would require preserving, or are things referenced by key.

mogoodrich:
Thanks @jmbabazi  @mseaton ... unfortunately I finished my review but then went back and made some tweaks so flagging you again...

mirebalais-modules/mirebalais_reporting/templates/percona-openmrs-db-restore.sh.erb
mseaton: reporting_report_request does not have foreign keys to any reports - it references everything by uuid

mirebalais-modules/mirebalais_reporting/templates/percona-openmrs-db-restore.sh.erb
mseaton: Ok, if that's true, that's fine.  As long as that behavior happens.

jmbabazi: Thanks everyone! yes the dump drops and re-creates the table

mseaton (approved):
Looks good James, thanks

PR #222, opened 2022-10-26
UHM-6790: Support configuring authentication module via Puppet

mseaton:


mirebalais-modules/openmrs/templates/openmrs-runtime.properties.erb
```diff
@@ -48,3 +48,16 @@ mail.smtp.ssl.protocols=TLSv1.2
 mail.transport.protocol=smtp
 mail.user=<%= @openmrs_mail_user %>
 mail.smtp.socketFactory.class=javax.net.ssl.SSLSocketFactory
+
+# Authentication settings
```
mogoodrich, line 52: Maybe add something to this comment noting that if authentication.scheme is blank it defaults to the "basic" scheme?

Just something so that people don't think that 2fa and/or the secret schemes are enabled just by having them in this file?

mogoodrich (approved):
One comment but lgtm...

mirebalais-modules/openmrs/templates/openmrs-runtime.properties.erb
mseaton: Done

PR #225, opened 2022-12-14
Uhm 6881 enable activity log

jmbabazi:


mseaton (approved):
This seems fine for now, but we should be careful about using this anywhere beyond data kind as we may inadvertently lose expected logging configuration on other systems.

PR #228, opened 2023-01-10
remove references to environment var that is no longer used

mogoodrich:
Pretty sure we no longer use this, so just wanted to clean up before we forgot... thoughts @jmbabazi ?


mseaton (approved):
LGTM assuming no one uses this by passing an argumnet into gem-update.

jmbabazi (approved):

PR #233, opened 2023-02-27
Draft - refactoring apzu.pp

jmbabazi:


jmbabazi (commented):
pwa_enable is need because we use this in apache default-ssl.conf file. We could change this approach if needed

mirebalais-modules/openmrs/manifests/openmrs_config_files.pp
```diff
@@ -0,0 +1,40 @@
...
+  file { "${tomcat_home_dir}/.OpenMRS/warehouse-connection.properties":
+    ensure  => present,
+    content => template('openmrs/warehouse-connection.properties.erb'),
+    owner   => $tomcat,
+    group   => $tomcat,
+    mode    => '0600',
+    require => File["${tomcat_home_dir}/.OpenMRS"]
+  }
+
+  file { "${tomcat_home_dir}/.OpenMRS/modules":
```
mogoodrich, line 33: This probably goes in the openmrs_malawi_distribution, as it's not really a configuration file, but just setting up the modules folder?  I don't think we set it up in the pihemr, it's probably handled by the debian package install?

jmbabazi: thanks @mogoodrich , the debian package installs /home/tomcat7 and the below. This is only for the pihemr and not for the malawi distribution. What are your thoughts on having a class that installs this for the malawi distribution and the rwanda distribution  (instead of having this in openmrs_malawi_distribution class)?

.OpenMRS 
   |-modules
  ├── openmrs-runtime.properties
  └── owa
    ├── labworkflow.zip
    └── orderentry.zi

jmbabazi (commented):
This class installs the openmrs home directory and runtime property. Useful for Malawi (and Rwanda). Maybe we should move the warehouse.properties files to another class since its fro Malawi only. 
@mseaton @mogoodrich @cioan

mseaton:
Yes, the warehouse.properties should be something that is only in Malawi

mirebalais-modules/openmrs/manifests/download_artifacts_from_maven_repo.pp
```diff
@@ -0,0 +1,60 @@
...
+  # dowload new distribution from maven repo
+  maven { "/tmp/${package_name}-${package_version}.zip":
+    groupid => "${maven_download_repo_package_group_id}",
+    artifactid => "${package_name}",
+    version => "${package_version}",
+    ensure => latest,
+    packaging => zip,
+    repos => "${maven_download_repo}",
+    require => [Exec['cleanup-openmrs-maven-distribution-dir'], Package['maven']],
+  }
```
mseaton, line 30: Ah, so you found a maven command in Puppet?  That's perfect.

mirebalais-modules/openmrs/manifests/download_artifacts_from_maven_repo.pp
```diff
@@ -0,0 +1,60 @@
...
+
+  # extract the new
+  exec { 'extract-distro':
+    command => "unzip -o /tmp/${package_name}-${package_version}.zip -d /tmp",
+    require => [ Exec['cleanup-openmrs-maven-distribution-dir'], Maven["/tmp/${package_name}-${package_version}.zip"], Package['unzip']]
+  }
+
```
mseaton, line 37: I still think that everything above this line should go into "install-distribution-from-maven.pp" and everything below this line should go in "install-config-from-maven.pp".  Keeping them separate is a bit cleaner, and would allow us to one day choose to re-use the "install-config" one in other jobs, even if the distro isn't installed from maven.

hieradata/neno-ci.pih-emr.org.yaml
```diff
@@ -6,6 +6,13 @@ server_timezone: Africa/Blantyre
 package_name: pihmalawi
 package_version: 7.0.0-SNAPSHOT
 
+config_name: "%{hiera('package_name')}-config"
+config_version: "%{hiera('package_version')}"
+
+maven_download_repo: http://mavenrepo.openmrs.org/nexus/content/repositories/public
+maven_download_repo_config_group_id: org.openmrs.module
+maven_download_repo_package_group_id: org.openmrs.distro
+
```
mseaton, line 15: It is possible that we'd have the distro and the config in different maven repositories too, but that's easy enough to change when the time comes too I guess

mseaton (commented):
Looks generally good @jmbabazi .  A few comments, but nothing critical.

mseaton (approved):
Quickly reviewing this looks pretty good to me @jmbabazi .  Let's give it a shot

mseaton (approved):
Looks good to me @jmbabazi

PR #235, opened 2023-05-09
Uhm 7187 timezone setting for dbevent

jmbabazi:


mirebalais-modules/mysql_setup/manifests/init.pp
```diff
@@ -128,11 +129,37 @@
...
+    command => "/usr/bin/mysql_tzinfo_to_sql /usr/share/zoneinfo | mysql -uroot -p${root_password} mysql",
+    user => root,
+    require => [Exec['set-root-password'], Exec['confirm-root-password'], Package['mysql-client-5.6'], Service[mysqld]]
+  }
+
+  # this is only added once and only if the variable doesn't exist
+  exec { 'add-timezone-if-not-exist':
+    command => "/bin/bash -c \"if ! grep -q 'default-time-zone=${timezone}' /etc/mysql/my.cnf; then sed -i '/^#timezone/a default-time-zone=${timezone}' /etc/mysql/my.cnf; fi\"",
+    require => [ File['/etc/mysql/my.cnf'], Service[mysqld], Exec['configure-timezone']],
+  }
```
mseaton, line 156: I don't understand the need for this line - why not just add this directly to my.cnf.erb (see my comment below)?

mirebalais-modules/mysql_setup/templates/my.cnf.erb
```diff
@@ -135,6 +135,8 @@ net_write_timeout = <%= @mysql_net_write_timeout %>
 wait_timeout = <%= @mysql_wait_timeout %>
 interactive_timeout = <%= @mysql_interactive_timeout %>
 
+# timezone
```
mseaton, line 138: Why wouldn't you just add `default-time-zone = <%= @timzone %>` here, rather than using that sed command above??

mseaton (changes requested):

mirebalais-modules/mysql_setup/templates/my.cnf.erb
jmbabazi: This was my first approach @mseaton but mysql would not start.

mirebalais-modules/mysql_setup/manifests/init.pp
```diff
@@ -128,11 +129,37 @@
...
     enable  => true,
     require => [ File['/etc/mysql/my.cnf'], File['/etc/mysql/my.cnf.fallback'], File['root_user_my.cnf'], Package['mysql-server-5.6'] ],
   }
+
+  ## these next blocks configures timezone in mysql
+  exec { 'configure-timezone':
+    command => "/usr/bin/mysql_tzinfo_to_sql /usr/share/zoneinfo | mysql -uroot -p${root_password} mysql",
+    user => root,
+    require => [Exec['set-root-password'], Exec['confirm-root-password'], Package['mysql-client-5.6'], Service[mysqld]]
+  }
```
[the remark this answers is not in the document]
jmbabazi: @mseaton , I changed the order.

mirebalais-modules/mysql_setup/manifests/init.pp
jmbabazi: @mseaton , we need this line because if we add it directly to my.cnf then mysql fails to start. 

If you re-call our testing yesterday, the workflow that worked was
- run /usr/bin/mysql_tzinfo_to_sql /usr/share/zoneinfo | mysql -uroot -p${root_password} mysql
- add default-time-zone=${timezone} to my.cnf
- restart mysql

mirebalais-modules/mysql_setup/manifests/init.pp
```diff
@@ -134,5 +135,25 @@
...
+    command => "/bin/bash -c \"if ! grep -q 'default-time-zone=${timezone}' /etc/mysql/my.cnf; then sed -i '/^#timezone/a default-time-zone=${timezone}' /etc/mysql/my.cnf; fi\"",
+    require => [ File['/etc/mysql/my.cnf'], Service[mysqld], Exec['configure-timezone']],
+  }
+
+  # restart mysql
+  exec { 'mysql-restart':
+    command     => "service mysql restart",
+    user        => 'root',
+    require     =>  [Service[mysqld], Exec['configure-timezone']]
+  }
```
mseaton, line 157: @jmbabazi I think this task needs to require Exec['add-timezone-if-not-exist'].

PR #240, opened 2023-07-14
UHM-7277: (reduce number of days of backups saved)

mogoodrich:


mseaton (approved):
LGTM, makes sense @mogoodrich

PR #241, opened 2023-07-14
UHM-7278: Enable Percona-based backups for Haiti production servers (…

mogoodrich:
…Haiti HIV)

### DIFF
--- a/hieradata/hai-cloud-inf-omrshiv.yaml
+++ b/hieradata/hai-cloud-inf-omrshiv.yaml
@@ -3,10 +3,11 @@ webapp_name: openmrs
 site_domain: haiti.pih-emr.org
 
 package_release: 'stable'
+package_version: '1.0.20221201210023-1'
 
 pih_config: haiti,haiti-hiv
 config_name: openmrs-config-zl
-config_version: 1.44.0
+config_version: 1.42.0
 
 frontend_name: "openmrs-frontend-zl"
 frontend_version: 1.3.0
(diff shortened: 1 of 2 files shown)


manifests/site.pp
```diff
@@ -103,6 +103,7 @@
   include mysql_setup
   include apache2
   include tomcat
+  include docker
```
mogoodrich, line 106: @mseaton  your steps *didn't* include including the "percona" node, but it looks like this is because it's not needed if Percona in run via docker, lmk if that's incorrect

hieradata/hai-cloud-inf-omrshiv.yaml
```diff
@@ -3,10 +3,11 @@ webapp_name: openmrs
 site_domain: haiti.pih-emr.org
 
 package_release: 'stable'
+package_version: '1.0.20221201210023-1'
 
 pih_config: haiti,haiti-hiv
 config_name: openmrs-config-zl
-config_version: 1.44.0
+config_version: 1.42.0
```
mogoodrich, line 10: Locking on the config and gold versions installed on Haiti HIV as noted here:

https://pihemr.atlassian.net/wiki/spaces/PIHEMR/pages/1562378273/PIH+EMR+Code+and+Config+Releases+and+Deployments
(I also confirmed via "dpkg -s pihemr" that this was the version installed on Haiti HIV

manifests/site.pp
mseaton: That's the idea @mogoodrich .  We'll see...

mseaton (approved):

PR #247, opened 2023-08-01
UHM-7310 - Update PETL test server to reflect production setup

mseaton:


mirebalais-modules/percona/templates/refresh-report-dbs-petl-test.sh.erb
```diff
@@ -0,0 +1,19 @@
...
+# Note, the 127.0.0.1 prefix on the port results in access being restricted to localhost
+# This allows the database user to be allowed access from '%' while still locking down access to the DB from outside
+
+restoreSite haiti/cange haiti_cange_mysql 127.0.0.1:3310
+restoreSite haiti/hinche haiti_hinche_mysql 127.0.0.1:3311
+restoreSite haiti/hsn haiti_hsn_mysql 127.0.0.1:3312
+restoreSite haiti/lacolline haiti_lacolline_mysql 127.0.0.1:3313
+restoreSite haiti/thomonde haiti_thomonde_mysql 127.0.0.1:3314
+restoreSite haiti/haitihiv haiti_haitihiv_mysql 127.0.0.1:3315
+restoreSite haiti/mirebalais haiti_mirebalais_mysql 127.0.0.1:3316
```
mseaton, line 19: Will add in the ces dbs to restore to this file once those are in place so petl-test can be used for testing both ZL and CES pipelines.

mseaton:
@mogoodrich I merged this in but would be good for you to look it over and sanity check and also get a sense of this new approach I took and what you think of it...

manifests/site.pp
```diff
@@ -632,13 +629,40 @@
...
+    reporting_system => 'petl-test'
+  }
+
+  petl::install { 'install-petl-zl-etl':
+    petl => "petl",
+    petl_user => "petl",
+    petl_home_dir => "/opt/petl",
+    petl_site => "zl-test",
+    petl_config_name => "zl-etl",
+    petl_config_version => "1.8.0-SNAPSHOT",
```
mogoodrich, line 648: It's slightly annoying that we now need to config these hear, outside of yaml files, but, yeah, thinking it through this is cleaner than having to petl_zl_site and petl_ces_site and having to have the petl module handle within it's init setting up multiple instances of Petl...  so anyway, just talking through this, no action here... :)

manifests/site.pp
```diff
@@ -632,13 +629,40 @@
...
+    petl_server_port => 9109,
```
mogoodrich, line 649: And side note, yeah, we should really clean up (and hopefully simplify) this file.

mogoodrich (commented):
LGTM, thanks @mseaton !

PR #249, opened 2023-08-15
Tomcat9

mogoodrich:


hieradata/common.yaml
```diff
@@ -50,9 +50,8 @@ backup_db_password: foo
 openmrs_percona_backup_enabled: false
 openmrs_percona_install_mode: ''
 backup_time: date +%Y%m%d-%H%M%S
-tomcat: tomcat7
```
mogoodrich, line 53: I went back and forth on this a bit, but I ended up removing the "tomcat" variable in favor of hardcoding "tomcat9".  My reasoning... although there are many (if not most) places where a direct swap out of "tomcat7" to "tomcat9" works, it shouldn't be assumed, and you can't install "tomcat7" just by setting this to "tomcat7", nor upgrade just by changing this to "tomcat9".  Seemed better to hardcode to 'tomcat9', so that when we upgrade to a higher version of tomcat it's easy (and forces us to) search on 'tomcat9' and analyze/tweak every place it is used.

hieradata/common.yaml
```diff
@@ -50,9 +50,8 @@ backup_db_password: foo
...
-tomcat_home_dir: /home/tomcat7
-tomcat_webapp_dir: /var/lib/tomcat7/webapps
+tomcat_home_dir: /home/tomcat
```
mogoodrich, line 53: That being said, I did switch the home dir from "tomcat7" to "tomcat" (and switched the user that runs tomcat from "tomcat7" to "tomcat", as this appears to be the standard practice).  This should allow us to keep this directory constant when we do further tomcat upgrades.

hieradata/common.yaml
```diff
@@ -50,9 +50,8 @@ backup_db_password: foo
...
+tomcat_webapp_dir: /var/lib/tomcat9/webapps
```
mogoodrich, line 54: This is the default webapp directory for Tomcat9

hieradata/common.yaml
```diff
@@ -219,7 +218,7 @@ terms_and_conditions_enabled: false
 ocl_package_url: ""
 
 # atomfeed
-openmrs_modules_dir: /home/tomcat7/.OpenMRS/modules
+openmrs_modules_dir: /home/tomcat/.OpenMRS/modules
 atomfeed_version: 1.0.13-SNAPSHOT
 atomfeed_repo_url: https://openmrs.jfrog.io/artifactory/snapshots
```
mogoodrich, line 223: (below: I just got rid of the entire bahmni project as it was outdated and no longer used)

mirebalais-modules/apache2/manifests/init.pp
```diff
@@ -1,5 +1,4 @@
 class apache2 (
-  $tomcat = hiera('tomcat'),
```
mogoodrich, line 2: first of many cases where I just remove this variable (it didn't even appear to be used in this file, fwiw)

mirebalais-modules/mirebalais_reporting/templates/archive/restart-server.sh.erb
```diff
@@ -7,7 +7,7 @@
 
 stopTomcat()
 	{
-	/etc/init.d/tomcat7 stop
+	service tomcat9 stop
```
mogoodrich, line 10: Not sure if there was a reason we we using the "/etc/init.d" instead of "service", I changed this to the standard (and it didn't appear that "/etc/init.d/tomcat9" had been set up in the new install)

mirebalais-modules/openmrs/manifests/atomfeed.pp
```diff
@@ -13,10 +12,10 @@
 
   file { "atomfeed":
     path => "$openmrs_modules_dir/atomfeed-${atomfeed_version}.omod",
-    owner   => $tomcat,
-    group   => $tomcat,
+    owner   => 'tomcat',
```
mogoodrich, line 15: first of many cases where I switch to hardcode to a (newly created) "tomcat" user, instead of a user based on the tomcat varaible

mirebalais-modules/openmrs/manifests/initial_setup.pp
```diff
@@ -32,8 +31,7 @@
     privileges => ['ALL'],
     table => '*.*',
     user => "root@localhost",
-    require => [Service['mysqld'],  Package[$package_name]],
-    notify  => Exec['set up base schema'];
```
mogoodrich, line 36: tangential, but I don't think this was needed

mirebalais-modules/openmrs/manifests/initial_setup.pp
```diff
@@ -81,11 +80,4 @@
     notify      => Exec['tomcat-restart']
   }
 
-  # hack to let us remote the mirebalais metadata module from the build; can be removed after it has been removed from all servers
-  exec { 'make mirebalais metadata module not mandatory - error running this command can be ignored when provisioning new server':
-    command     => "mysql -u${openmrs_db_user} -p'${openmrs_db_password}' ${openmrs_db} -e 'update ${openmrs_db}.global_property set property_value=\"false\" where property=\"mirebalaismetadata.mandatory\"'",
-    user        => 'root',
-    require => Package[$package_name],
-  }
-
```
mogoodrich, line 90: Detritus clean-up

mirebalais-modules/percona/files/percona-restore.sh
```diff
@@ -169,7 +169,7 @@ fi
 
 if [ "${RESTART_OPENMRS}" == "true" ]; then
   echoWithDate "Stopping Tomcat"
-  /etc/init.d/${TOMCAT_USER} stop
+  service tomcat9 stop
```
mogoodrich, line 172: see previous comment about moving away from "/etc/init."

mirebalais-modules/tomcat/files/logging.properties
```diff
@@ -13,46 +13,46 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-handlers = 1catalina.org.apache.juli.FileHandler, 2localhost.org.apache.juli.Fil
-eHandler, java.util.logging.ConsoleHandler
+handlers = 1catalina.org.apache.juli.AsyncFileHandler, 2localhost.org.apache.juli.AsyncFileHandler, java.util.logging.ConsoleHandler
```
mogoodrich, line 16: I basically changed this file to the default installed with tomcat9, with one change noted below

mirebalais-modules/tomcat/files/logging.properties
```diff
@@ -13,46 +13,46 @@
...
 # Handler specific properties.
 # Describes specific configuration info for Handlers.
 ############################################################
 
-# see http://tomcat.apache.org/tomcat-6.0-doc/api/org/apache/juli/FileHandler.html
-# we set rotatable = false because we use log4j to handle logging
+1catalina.org.apache.juli.AsyncFileHandler.level = FINE
+1catalina.org.apache.juli.AsyncFileHandler.directory = ${catalina.base}/logs
+1catalina.org.apache.juli.AsyncFileHandler.prefix = catalina.
+1catalina.org.apache.juli.AsyncFileHandler.rotatable = false
```
mogoodrich, line 28: I added this line and removed one that appeared to set log rotation to 90 days... assumedly we turn this off and defer to logrotate, but we should confirm this works as expected.

mirebalais-modules/tomcat/files/logrotate
```diff
@@ -1,37 +1,37 @@
-/var/lib/tomcat7/logs/catalina.out  {
+/var/lib/tomcat9/logs/catalina.out  {
```
mogoodrich, line 1: should test that this is working after install.... I didn't run into any errors, but didn't run for a day either to confirm it starts rotating...

mirebalais-modules/tomcat/manifests/init.pp
```diff
@@ -9,133 +8,101 @@
     $enable_http_8080 = hiera('tomcat_enable_http_8080')
   ){
 
-  # TODO--remove these three purges after tomcat7 apt-get update has been installed on all servers
```
mogoodrich, line 12: No where is running tomcat6, so I could remove all this old stuff.

mirebalais-modules/tomcat/manifests/init.pp
```diff
@@ -9,133 +8,101 @@
...
-  }
 
-  file { "/usr/local/apache-tomcat-7.0.68" :
-    ensure => absent,
-    recurse => true,
-    purge => true,
-    force => true,
+
+  package { 'tomcat7':
+    ensure => purged
```
mogoodrich, line 15: Purge old tomcat7

mirebalais-modules/tomcat/manifests/init.pp
```diff
@@ -9,133 +8,101 @@
...
   }
 
-  file { "/usr/local/$tomcat" :
-    ensure => absent,
-    recurse => true,
-    purge => true,
-    force => true,
+  # TODO clean up old tomcat directory and user
```
mogoodrich, line 18: I didn't add anything (yet) to clean out the old "/var/lib/tomcat7" directory... seemed worth keeping it around until we confirm everything is working with tomcat9

mirebalais-modules/tomcat/manifests/init.pp
```diff
@@ -9,133 +8,101 @@
...
+
+  user { 'tomcat':
+    ensure => 'present',
+    home   => "${tomcat_home_dir}",
+    shell  => '/bin/sh',
```
mogoodrich, line 23: Create the new tomcat user and home dir.

mirebalais-modules/tomcat/manifests/init.pp
```diff
@@ -9,133 +8,101 @@
...
 
-  exec { 'remove /etc/default/tomcat':
-    command     => "rm /etc/default/$tomcat",
-    subscribe   => [ File["/usr/local/apache-tomcat-6.0.36"], File["/usr/local/apache-tomcat-7.0.62"], File["/usr/local/apache-tomcat-7.0.68"]],
-    refreshonly => true,
+  exec { 'move-openmrs-home-directory':
+    command => "mv /home/tomcat7/.OpenMRS ${tomcat_home_dir}",
+    onlyif  => "test -d /home/tomcat7/.OpenMRS",
+    require => File[$tomcat_home_dir],
+    notify => Exec['change-home-directory-permissions']
```
mogoodrich, line 38: Does rename of "/home/tomcat7" to "/home/tomcat"

mirebalais-modules/tomcat/manifests/init.pp
```diff
@@ -9,133 +8,101 @@
...
   }
 
-  exec { 'remove /etc/logrotate.d/tomcat':
-    command     => "rm /etc/logrotate.d/$tomcat",
-    subscribe   => [ File["/usr/local/apache-tomcat-6.0.36"], File["/usr/local/apache-tomcat-7.0.62"], File["/usr/local/apache-tomcat-7.0.68"]],
+  exec { 'change-home-directory-permissions':
+    command => "chown -R tomcat ${tomcat_home_dir}/.OpenMRS && chgrp -R tomcat ${tomcat_home_dir}/.OpenMRS",
+    require => Exec['move-openmrs-home-directory'],
```
mogoodrich, line 43: Update permissions after the move

mirebalais-modules/tomcat/manifests/init.pp
```diff
@@ -9,133 +8,101 @@
...
 
   # install the proper version of tomcat via apt-get
-  package { $tomcat :
+  package { 'tomcat9' :
     ensure => installed,
-    require => [ User[$tomcat], File["/usr/local/apache-tomcat-6.0.36"], File["/usr/local/apache-tomcat-7.0.62"], File["/usr/local/apache-tomcat-7.0.68"],
-        Exec['remove /etc/init.d/tomcat'], Exec['remove /etc/default/tomcat'], Exec['remove /etc/logrotate.d/tomcat'], Package['openjdk-8-jdk']],
-    notify  => Service["$tomcat"]
+    require => [ Package['openjdk-8-jdk'], Package['tomcat7'], User['tomcat']],
+    notify  => Service['tomcat9']
```
mogoodrich, line 51: install the new Tomcat 9

mirebalais-modules/tomcat/templates/default.erb
```diff
@@ -1,10 +1,3 @@
-# Run Tomcat as this user ID. Not setting this or leaving it blank will use the
-# default of tomcat7.
-TOMCAT7_USER=tomcat7
-
-# Run Tomcat as this group ID. Not setting this or leaving it blank will use
-# the default of tomcat7.
-TOMCAT7_GROUP=tomcat7
```
mogoodrich, line 7: I took the default "default" file from the Tomcat9 install... specifically, no longer have this (custom?) stuff that ran Tomcat using the "tomcat7" (it now uses a new "tomcat" user)

mirebalais-modules/tomcat/templates/server.xml.erb
```diff
@@ -25,8 +25,6 @@
   <Listener className="org.apache.catalina.security.SecurityListener" />
   <!--APR library loader. Documentation at /docs/apr.html -->
   <Listener className="org.apache.catalina.core.AprLifecycleListener" SSLEngine="on" />
-  <!--Initialize Jasper prior to webapps are loaded. Documentation at /docs/jasper-howto.html -->
-  <Listener className="org.apache.catalina.core.JasperListener" />
```
mogoodrich, line 29: Evidently this is no longer available/needed in Tomcat9

mirebalais-modules/tomcat/templates/server.xml.erb
```diff
@@ -94,7 +92,7 @@
 
     <!-- Define an AJP 1.3 Connector on port 8009 -->
     <Connector port="8009" address="127.0.0.1" protocol="AJP/1.3" redirectPort="8443"
-		URIEncoding="UTF-8"/>
+		URIEncoding="UTF-8" secretRequired="false" />
```
mogoodrich, line 95: Good reminder to see this... evidently it's best practice to use a "secret" with the connector, and, starting with Tomcat9 you have to explicitly disable it.   We should exploring this further and see if this is something we want to set up.

mogoodrich:
Ready for review... however, definitely don't merge!  My plan is to test on a single server first (likely HUM-CI), see ticket for more information.

hieradata/common.yaml
mseaton: Quickly googling this, I wonder if it wouldn't be better to use a standard directory for the webapps that doesn't need to change over time.  For example, if we put the webapps at `/opt/openmrs/webapps`, then we can tell Tomcat to use that directory by configuring this in server.xml:

```<Host name="localhost" appBase="/opt/openmrs/webapps" unpackWARs="true" autoDeploy="true">```

So, as long as we have some variable in our Puppet config that defines exactly where we want the webapps to go, we should be able to both a) ensure that directory exists, and b) update the server.xml.erb file to use this variable to set the appbase.  This would seem like a better approach, and would mean we don't need to update our Debian package again for the next Tomcat version.

File: mirebalais-modules/logging/templates/logstash.conf.erb [hunk 24]
```diff
@@ -88,7 +88,7 @@ input {
  }
 
  file {
-  path => ["/usr/local/<%= @tomcat %>/logs/catalina.out"]
+  path => ["/usr/local/tomcat9/logs/catalina.out"]
```
mseaton, line 91: I _believe_ if we set a CATALINA_OUT environment variable, then this can point to any arbitrary path, so if we want to configure this to go somewhere else and not have it have "tomcat9" in the path, we should be able to do that.

mirebalais-modules/mirebalais_reporting/templates/archive/restart-server.sh.erb
mseaton: Seems better, let's go with it

mirebalais-modules/openmrs/manifests/config_files.pp
```diff
@@ -33,54 +32,55 @@
...
-      owner   => $tomcat,
-      group   => $tomcat,
+      owner   => 'tomcat',
+      group   => 'tomcat',
       mode    => '0644',
       require => File["${tomcat_home_dir}/.OpenMRS"]
     }
 
-    file { "/home/${tomcat}/.OpenMRS/frontend":
+    file { "/home/tomcat/.OpenMRS/frontend":
```
mseaton, line 56: Shouldn't this be `${tomcat_home_dir}/.OpenMRS/frontend` ?

mirebalais-modules/openmrs/manifests/initial_setup.pp
```diff
@@ -44,24 +42,25 @@
...
+    group   => 'tomcat',
     mode    => '0700',
-    require => File["/usr/local/liquibase.jar"]
+    require => [ Package['tomcat9'], File["/usr/local/liquibase.jar"] ]
   }
 
   exec { 'set up base schema':
     cwd         =>  '/usr/local/',
-    command     => "java -Dliquibase.databaseChangeLogTableName=liquibasechangelog -Dliquibase.databaseChangeLogLockTableName=liquibasechangeloglock -jar liquibase.jar --driver=com.mysql.jdbc.Driver --classpath=/var/lib/${tomcat}/webapps/${webapp_name}.war --url=jdbc:mysql://localhost:3306/${openmrs_db} --changeLogFile=liquibase-schema-only.xml --username=${openmrs_db_user} --password='${openmrs_db_password}' update",
+    command     => "java -Dliquibase.databaseChangeLogTableName=liquibasechangelog -Dliquibase.databaseChangeLogLockTableName=liquibasechangeloglock -jar liquibase.jar --driver=com.mysql.jdbc.Driver --classpath=/var/lib/tomcat9/webapps/${webapp_name}.war --url=jdbc:mysql://localhost:3306/${openmrs_db} --changeLogFile=liquibase-schema-only.xml --username=${openmrs_db_user} --password='${openmrs_db_password}' update",
```
mseaton, line 63: Note this would change if you go with what I put in my first comment, same with the below

mirebalais-modules/openmrs/manifests/initial_setup.pp
mseaton: Woot

mirebalais-modules/tomcat/manifests/init.pp
mseaton: Not really, it moves the .OpenMRS directory

mirebalais-modules/tomcat/manifests/init.pp
mseaton: I think you can just do this in one command `chown -R tomcat:tomcat ${tomcat_home_dir}/.OpenMRS`

mseaton (commented):
Couple of comments, nothing major.  All looks good to me.  Having seen the finished product, I think I probably might have worked better to have and use the following variables:

$tomcat_user
$tomcat_group
$tomcat_home
$tomcat_webapps_dir
$tomcat_logs_dir
$tomcat_service

And then used those variables throughout the tasks and templates.  I think this will likely allow things to stay consistent in the future.  But hard-coding works fine too...

mirebalais-modules/tomcat/manifests/init.pp
mogoodrich: Oh right, good point, will change.

mirebalais-modules/openmrs/manifests/config_files.pp
mogoodrich: Yep, good catch.

mirebalais-modules/tomcat/manifests/init.pp
mogoodrich: oh, right, fair, and that was the intent.

PR #250, opened 2023-09-25
Uhm 7371 Create ACME user

jmgihana:


hieradata/common.yaml
```diff
@@ -200,7 +201,7 @@ apache_cron_restart_min: 03
 az_secret: ENC[1oQfg4Gi51CDU3XTOIkHatJn2W+vAxx1Juj/wYnVwyRtEFk7hm7XAkA0CbKlsUcUdED4rBiSDIYAlTiO3aBryg==]
 az_url: ENC[6NweILo1Eq7pKNo88HD1o/fyAWOijtE/7o3Pn+pd8tJ9q2Z1PerwvhtXMsVi2nl37EjQL96k8gApwQCE+nKnaw==]
 az_backup_folder_path: ''
-#az_backup_hour: '*/7'
+#u;llaz_backup_hour: '*/7'
```
mseaton, line 204: This looks like an unintentional typo?

mseaton (commented):
Just did a cursory review, defer to @mogoodrich .  But found one typo I think

mogoodrich:
Thanks @jmgihana , will review later today, but one question... did we confirm this both will update/clean up an existing server that already is running it as root, as well as a fresh install?

jmgihana:
> Thanks @jmgihana , will review later today, but one question... did we confirm this both will update/clean up an existing server that already is running it as root, as well as a fresh install?

I've added a block that should delete this Cron for user root. Good idea to test this on a fresh install.

hieradata/common.yaml
```diff
@@ -187,6 +187,7 @@ ssl_chain_file: 'fullchain.pem'
 ssl_key_dir: "/etc/letsencrypt/live/%{hiera('site_domain')}"
 ssl_key_file: 'privkey.pem'
 ssl_use_letsencrypt: true
+acme_user: ENC[2tXA1UMPe7zSGFq27HtWPg==:3FFCUm2Ldb+3E5OuuQeecg==:RXhzZHhCSDZYWGdHdGJCSnFYK282UT09]
```
mogoodrich, line 190: As long as this is the value I think it is... :)

mogoodrich (approved):
Thanks @jmgihana , sorry for the delayed review but looks good to me!

PR #252, opened 2023-11-02
UHM-7501: Create new LetsEncrypt SSL certificates

mogoodrich:


hieradata/common.yaml
```diff
@@ -187,6 +187,7 @@ ssl_chain_file: 'fullchain.pem'
 ssl_key_dir: "/etc/letsencrypt/live/%{hiera('site_domain')}"
 ssl_key_file: 'privkey.pem'
 ssl_use_letsencrypt: true
+acme_version: 3.0.7
```
mogoodrich, line 190: locks on a version of acme to use, previously we were always checking out the "master" branch on every deploy

mirebalais-modules/apache2/manifests/init.pp
```diff
@@ -130,16 +133,19 @@
     require => File["/var/$acme_user"]
   }
 
+  # create all the directories where we install the acme scripts
   file { "/etc/letsencrypt" :
     ensure => directory,
     owner   => "$acme_user",
     group   => "$acme_user",
+    recurse => true,
```
mogoodrich, line 141: Need this to set directories/files created previously using root to be owned by acme

mirebalais-modules/apache2/manifests/init.pp
```diff
@@ -148,46 +154,63 @@
     owner   => "$acme_user",
     group   => "$acme_user",
     mode    => '0710',
+    recurse => true,
     require => File["/etc/letsencrypt/live"]
   }
 
-  file { "install-letsencrypt.sh":
+  # install install-certs.sh
+  file { "install-certs.sh":
```
mogoodrich, line 162: changed the name to install-certs, which he had been using in other cases, and is clearer

mirebalais-modules/apache2/manifests/init.pp
```diff
@@ -148,46 +154,63 @@
...
     require => File["/var/$acme_user"],
-    notify => Exec['run install letsencrypt'],
+    notify => [ Exec['run install certs'], Exec['install acme.sh tool'] ],
   }
 
-  exec { "download acme from the git repo":
-    command => "rm -rf /var/$acme_user/acme.sh && git clone https://github.com/acmesh-official/acme.sh.git /var/$acme_user/acme.sh && chown -R $acme_user:$acme_user /var/$acme_user/acme.sh",
-    require => File["/var/$acme_user"]
+  # install acme
+  # note: refresh-only, so this only runs when the install cert script changes (see notify above)
```
mogoodrich, line 174: note comment ^^

mirebalais-modules/apache2/manifests/init.pp
```diff
@@ -148,46 +154,63 @@
...
+  exec { "install acme.sh tool":
+    command => "sudo -H -u acme bash -c 'wget -O -  https://raw.githubusercontent.com/acmesh-official/acme.sh/$acme_version/acme.sh | sh -s -- --install-online -m  emrsysadmin@pih.org --home /var/acme'",
```
mogoodrich, line 176: note that we now install a specific acme version, and insteadl of using git to checkout, we use another recommended approach that just does a wget and runs the "install" upon download.  Also note that we run as the acme user.

mirebalais-modules/apache2/templates/install-certs.sh.erb
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+# Note that this script is intended to be run as the acme user
+
+set -x
+
+# a bit of hack/trick... this does nothing, but assures that each time we upgrade acme_version this script changes,
+# and any changes to this script trigger a re-install of acme.sh (see apache/init.pp), so this triggers a acme.sh reinstall on a version change
+echo "using acme version <%= @acme_version %>"
```
mogoodrich, line 9: note comment above ^^

mogoodrich (commented):
@cioan  @mseaton  ready for your review.

As a meta-point, although "best" practice for testing Puppet would be to test it all out on a Vagrant box, admittedly we rarely do that.

Instead of doing a total trial-and-error approach on the master branch, I've started doing the following:

* develop in a feature branch, as normal
* pick a CI server you are going to test on and disable it in Bamboo and then in /etc/puppet check out your feature branch
* iterate and test on that server, pushing/pulling changes as needed, and then open a PR when ready.

mirebalais-modules/apache2/templates/install-certs.sh.erb
mseaton: Smaat

mseaton (approved):
LGTM, nice work

cioan (approved):
Looks good! Thanks @mogoodrich !

mogoodrich:
Thanks @cioan  @mseaton

PR #255, opened 2023-11-14
UHM-7562 - Create custom debian.cnf for mysql restarts

mseaton:


mogoodrich (approved):
Okay, so I think I get it... we just set up a my.cnf for this "debian user" that tells it to connect to mysql via the root user and password?  Makes sense to me.

PR #256, opened 2023-11-14
UHM-7570 - Update azure authentication in puppet

mseaton:


mogoodrich (approved):
From what I'm reading, it does seem like the whole set of parameters comprise the shared access signatures (SAS) token, and therefore likely makes sense to encrypt it.

PR #258, opened 2023-12-20
UHM-7643: Change Sierra Leone Backup Folder Path

mogoodrich:


mseaton (approved):
LGTM.  Would be nice if we could clean up the historical files and move them into the new folders, but not really sure if that is very straightforward.

PR #262, opened 2024-02-07
UHM-7717 - Enable percona backups on wellbody and kgh servers

mseaton:


mogoodrich (commented):
LGTM, thanks @mseaton

PR #263, opened 2024-02-29
UHM-7761 - Add maintenance for cleaning up disk space.

mseaton:


mirebalais-modules/maintenance/files/reclaim-disk-space.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+#
+# Script to facilitate cleaning up disk space
+
```
mogoodrich, line 4: Can we add a comment "Note that we don't want to run this script on on Production servers because it will prevent us from staging PIH EMR debian packages before install"

mogoodrich (approved):
LGTM, thanks @mseaton !  See my notes, I just want to add a comment so that we don't inadvertently install this on production servers because right now the production servers should stage new releases of the PIH EMR debian package automatically, and based on my testing, the apt-get clean would basically undo that.

mirebalais-modules/maintenance/files/reclaim-disk-space.sh
mseaton: I guess @mogoodrich , but that isn't really 100% true.  There is only a small window of time during which we don't want to run this on production servers, between when a gold is cut and when it is released.  The vast majority of the time this script would be safe and appropriate to run on any server on which we need to free up some space (though I'm not sure how much benefit it would have on production servers).  I'd be more inclined to put the comment in common.yml where this is set by default to not enable the cron.

mseaton: See comment I added above.

mogoodrich: Fair enough, that comment make sense.  I thought I had added a comment in the common.yaml as well, I'd be fine (and it would probably be better) if it was there instead.

I just think this is something we could easily forget and enable nightly on all the production servers, which, for instance would have made the recent KGH and Wellbody installs much more time consuming.

Tangent: when I was researching this, I did try to check if there was any "clean" parameter that would say "delete every version but the most recent download", but I didn't find it (though I didn't dig in super deep).

